PR #2, opened 2021-03-10
RFC 2: New quilt.mod.json Format Proposal

Haven-King:
This RFC describes a format to be used by mods to declare their mod.json files. This is a decision that needs to be decided on fairly early in the tech conversation, so I'm opening this now.

[Rendered view](https://github.com/QuiltMC/rfcs/blob/quilt.mod.json/loader/0002-quilt.mod.json.md)


loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,362 @@
...
+| Object | False    |
+
+Defines mods that this mod will not function without.
+
+A collection of `key: value` pairs, where each key is in the form of either `mavenGroup:modId` or `modId` and each value is either a [dependency object]() or an array of dependency objects. If an array of dependency objects is provided, the dependency matches if it matches ANY of the dependency objects.
+
+#### Dependency Object
+A dependency object is made up of a version specifier and a reason for that dependency. It may be represented in any of the following ways:
+```json5
+// As a JSON object, with both available fields
```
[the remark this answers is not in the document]
williambl: Making a mod that interfaces with as many others as it can, the lack of something like this is one of the few things that have bugged me about `fabric.mod.json`s.

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,362 @@
...
+* `^1.0.0` — Matches the most recent version greater than or equal to version 1.0.0 and less than 2.0.0
+
+### The `breaks` field
+| Type   | Required |
+|--------|----------|
+| Object | False    |
+
+Defines mods that this mod either breaks or is broken by.
+
+A collection of `key: value` pairs, where each key is in the form of either `mavenGroup:modId` or `modId` and each value is either a [dependency object]() or an array of dependency objects. If an array of dependency objects is provided, the dependency matches if it matches ANY of the dependency objects.
```
SpaceClouds42, line 169: It would be useful if this had an optional `requires` when using a dependency object to define a mod breakage. This could be used for the sodium/indium case.
```json5
breaks: {
  sodium: {
    version: "*",
    reason: "Incompatible rendering mechanics",
    requires: "indium" // Could be a dependency object with version and/or use mavenGroup:modId
  }
}
```

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,362 @@
...
+    * [group_id](#the-group_id-field) — The Maven group id
+    * [mod_id](#the-mod_id-field) — The mod id
+    * [provides](#the-provides-field) — Alternative mod ids provided by this mod
+    * [version](#the-version-field) — The mods version
+    * [entrypoints](#the-entrypoints-field) — Collection of entrypoints
+    * [plugins](#the-plugins-field) — Collection of plugins
+    * [jars](#the-jars-field) — Array of nested JARs to be loaded
+    * [language_adapters](#the-language_adapters-field) — Array of language adapters
+    * [depends](#the-depends-field) — Collection of mod dependencies
+    * [breaks](#the-breaks-field) — Collection of mods that this mod is incompatible with
```
BeetMacol, line 21: Why are the dependency resolution types limited to `depends` and `breaks`?  
More types of the mod relations could be added and all of them could be grouped in a `relation` object.

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,362 @@
...
+    * [repositories](#the-repositories-field) — Array of maven repositories
+    * [metadata](#the-metadata-field) — Extra information about this mod and/or its authors
```
BeetMacol, line 23: How is the metadata related to loading? I think it should be moved outside the `quilt_loader` object.

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,362 @@
...
+* [access_widener](#the-access_widener-field) — Path(s) to accesswidener file(s)
+* [minecraft](#the-minecraft-field) - Minecraft related options
+    * [environment](#the-environment-field) — What game environments this mod applies to
+    
+## The `schema_version` field
+| Type   | Required |
+|--------|----------|
+| Number | True     |
+
+The quilt mod file schema version to be used for parsing this file. Currently, the only valid version is 1.
```
BeetMacol, line 40: Why is the first schema version `1` instead of `0`?

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,362 @@
...
+
+```json5
+// As a single JSON string, containing only the version specifier
+"*"
+```
+
+#### Version Specifier
+A version range specifier can make use of any of the following patterns:
+* `*` — Matches any version. Will fetch the latest version available if needed
+* `1.0.0` — Matches exactly version 1.0.0 and no other versions
```
Boundarybreaker, line 152: Rust's dependency management would interpret this as `^1.0.0` instead of `=1.00`, which I feel is a much more future-proof and robust way to handle unqualified version strings. If patch versions are used as intended, `1.0.1` should be api-compatible with anything targeting `1.0.0`. We should mirror this behavior.

loader/0002-quilt.mod.json.md
i509VCB: Version 0 in a way implies unreleased.

loader/0002-quilt.mod.json.md
i509VCB: There was some earlier discussion here - not sure if that answers any of your question?

https://github.com/QuiltMC/rfcs/pull/2#discussion_r591150398

Prospector: It was here. Personally I don't think it should be interpreted as `=`, since you should have to explicitly depend on an exact version, which should be an exceedingly rare case. https://github.com/QuiltMC/rfcs/pull/2#discussion_r590915393

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,362 @@
...
+}
+```
+The `"description"` field is optional.
+
+#### The `icon` field
+| Type          | Required |
+|---------------|----------|
+| Object/String | False    |
+
+One or more paths to a square .PNG file. If an object is provided, the keys must be the resolution of the corresponding file. For example:
```
NeRdTheNed, line 242: Is constraining the mod icon to only be a .PNG file necessary? [ImageIO on Java 8 natively supports at least 4 other image formats](https://docs.oracle.com/javase/8/docs/api/javax/imageio/package-summary.html) (more modern versions of Java support [an additional format](https://docs.oracle.com/en/java/javase/16/docs/api/java.desktop/javax/imageio/package-summary.html) on top of that), and has support for plugins which provide the ability to load other image formats. Additionally, would it be considered to be an error if a mod provides a non-square .PNG, or is handling this case left to whatever is parsing the quilt.mod.json file?

Boundarybreaker: The limitation to PNG is primarily because Minecraft doesn't super like binding textures for any other type of image, as far as I know.

Haven-King: I would maybe be willing to support SVG, but not much else. A non-square PNG breaks spec, so that would be considered an error.

Prospector: SVGs would put quite the burden on those who implement this imo, even for stuff like Mod Menu.

Haven-King: > SVGs would put quite the burden on those who implement this imo, even/especially for stuff like Mod Menu.

It certainly would, it's just the only other format I could justify including. PNG alone is fine.

0002-quilt.mod.json.md
```diff
@@ -0,0 +1,254 @@
...
+A collection of `"key": value` pairs, where each key is in the form of either `mavenGroup:modId` or `modId` and each value is either a version range specifier or array of version range specifiers. If an array of range specifiers is provided, the version matches if it matches ANY of the listed specifiers. A version range specifier can make use of any of the following patterns:
+* `*` — Matches any version. Will fetch the latest version available if needed
+* `=1.0.0` — Matches exactly version 1.0.0 and no other versions
+* `>=1.0.0` — Matches any version greater than or equal to 1.0.0
+* `>1.0.0` — Matches any version greater than version 1.0.0
+* `<=1.0.0` — Matches any version less than or equal to version 1.0.0
+* `<1.0.0` — Matches any version less than version 1.0.0
+* `1.0.x` — Matches any version with major version 1 and minor version 0.
+* `~1.0.0` — Matches the most recent version greater than or equal to version 1.0.0 and less than 1.1.0
+* `^1.0.0` — Matches the most recent version greater than or equal to version 1.0.0 and less than 2.0.0
```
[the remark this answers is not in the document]
Boundarybreaker: Rust treats unqualified as `^` because patch bumps aren't supposed to have major breaking changes. I guess my only concern would be Minecraft itself, ex. 1.16.2

loader/0002-quilt.mod.json.md
BeetMacol: I would leave it PNG only. There are a lot of mod icons that are SVG, but any SVG icon can be rendered to a high resolution PNG image.

Haven-King: The advantage to allowing SVG icons would be so that mods can have high resolution icons in-game without shipping large, high quality PNG's. I agree that it is probably best to forgo this, however.

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,362 @@
...
+    * [repositories](#the-repositories-field) — Array of maven repositories
+    * [metadata](#the-metadata-field) — Extra information about this mod and/or its authors
+        * [name](#the-name-field) — Human-readable name of this mod
+        * [description](#the-description-field) — Human-readable description of this mod
+        * [contributors](#the-contributors-field) — Collection of contributors to this mod
+        * [contact](#the-contact-field) — Collection of contact information
+        * [license](#the-license-field) — One or more licenses this project is under
+        * [icon](#the-icon-field) — The icon or icons associated with this project
+* [mixin](#the-mixin-field) — Path(s) to mixin config file(s)
+* [access_widener](#the-access_widener-field) — Path(s) to accesswidener file(s)
```
redcreeper14385, line 31: Surely `mixin` and `access_widener` should be within the loader object?

loader/0002-quilt.mod.json.md
redcreeper14385: I agree that it should be interpreted as `^1.0.0`, of course Minecraft (as mentioned elsewhere) could cause issues but this is up to the dev to specify properly.

loader/0002-quilt.mod.json.md
redcreeper14385: PNG only I would say.

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,362 @@
...
+```
+
+## Placeholders
+Strings with defined formats such as [the version field](#the-version-field) can instead supply a string matching the regex `^\$\{[a-zA-Z_$][a-zA-Z\d_$]*\}$` in a development environment to be replaced on build. The quilt-gradle plugin defines usage of the value `${version}` as a placeholder for the version declared in gradle.properties.
+
+# Drawbacks
+The primary drawback to this proposed format is the break from convention established by the Fabric project. It may make it more difficult for modders to adjust to the new toolchain if they are having to make drastic changes to their mod files.
+
+# Rationale and Alternatives
+It makes sense to take advantage of the breaking divergence of the toolchain to make changes that are better for Quilt's future. We could stick more strictly to the fabric.mod.json specification, but we would lose a lot of the advantages that this iteration has.
```
redcreeper14385, line 351: I think that compatibility will be lost at some point anyway, and it is worth sacrificing for better features.

redcreeper14385 (commented):
I have a few comments but mostly looks good.

0002-quilt.mod.json.md
```diff
@@ -0,0 +1,254 @@
...
+| String | True     |
+
+A unique identifier or the mod or library defined by this file, matching the `^[a-z][a-z0-9-_]{1,63}$` regular expression.
+
+### The `version` field
+| Type   | Required |
+|--------|----------|
+| String | True     |
+
+Must conform to the [Semantic Versioning 2.0.0 specification](https://semver.org/)
```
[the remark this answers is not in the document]
redcreeper14385: Something like `InvalidVersioningError`, causing a crash just like any other invalid field.

loader/0002-quilt.mod.json.md
Haven-King: Tentative plan is to have mixin and AW be phased out when ASMR becomes viable. For this reason, they will each be treated as loader plugins rather than internal parts of loader itself, though they won't actually *be* loader plugins until ASMR is available.

redcreeper14385: Makes sense 👍

redcreeper14385: While we're on this section of code, I have a question about `license` and the ability to specify multiple licenses? Would it be a matter of "pick one", and would they all need to be compatible? Would this be checked or is it up to the dev?

i509VCB: It's going to be very complicated to model and or or in json for this. Earlier discussion led to a conclusion that you should check the repo or license files yourself for exact terms.

redcreeper14385: That makes sense.

OroArmor:
While this spec is really comprehensive and thought out. There is a lot of data here that can be filled in and many people wouldn't know how to add. Some form of a program to make a json for this spec would be amazing. (I could make it, just would need a couple hours to get it done today)

BeetMacol:
> While this spec is really comprehensive and thought out. There is a lot of data here that can be filled in and many people wouldn't know how to add. Some form of a program to make a json for this spec would be amazing. (I could make it, just would need a couple hours to get it done today)

I think a wiki page + the file in the template mod would be enough, but such a program can be very handy for both the beginners and advanced modders.

OroArmor:
Yeah, I'm just saying that it would make complex mod json files way easier to verify

zml2008:
I fully expect that once this spec is finalized, someone will write a [json schema](https://json-schema.org) for the QMJ format, which will allow various tools to validate it.

kb-1000:
There already [is one for fabric.mod.json](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/fabric.mod.json), if QMJ files aren't too different I can probably derive something from it.

Scotsguy (approved):

loader/0002-quilt.mod.json.md
i509VCB: The way I see it the metadata for a mod specifically regards information needed to load the mod for quilt loader.

If you were to take the api surface route, quilt loader provides the apis to introspect this data so it makes sense here also.

loader/0002-quilt.mod.json.md
i509VCB: Just for some context - not asserting a decision here: the old `suggests` resolution type was never really used. `conflicts` type did print a warning but it was always really a soft resolution type.

loader/0002-quilt.mod.json.md
i509VCB: `requires` in the dependency object could be a bit confusing - implying sodium only breaks when indium is present.

`unless` would probably make it more clear.

Recursion could be a concern here but that is a thing for the dependency resolver to manage.

@AlexIIL I assume this wouldn't be too hard to model in the new partial-boolean solver. IF we consider this.

loader/0002-quilt.mod.json.md
BeetMacol: > The way I see it the metadata for a mod specifically regards information needed to load the mod for quilt loader.

Can't see the way it's associated.

> If you were to take the api surface route, quilt loader provides the apis to introspect this data so it makes sense here also.

The loader provides an API for any field in the `quilt.mod.json` file.

loader/0002-quilt.mod.json.md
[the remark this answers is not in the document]
wagyourtail: It might also be useful to add a "side" option, sometimes dependencies arn't needed on the 'server' side. the only option with the current stuff is to not add the mod to my depends list.

loader/0002-quilt.mod.json.md
AlexIIL: Yes, this would probably be fairly simple to add to mod resolving.

(Although this is *also* possible in the original solver - partial-boolean *just* allows us to inspect the cause of an error, it doesn't actually add any new functionality).

loader/0002-quilt.mod.json.md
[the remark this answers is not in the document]
AlexIIL: @wagyourtail I think that's part of the discussion above? (although swap `side` for `environment`) https://github.com/QuiltMC/rfcs/pull/2#discussion_r590935371

loader/0002-quilt.mod.json.md
Haven-King: I have changed the specification to behave like `^1.0.0`.

Haven-King:
I've just made large-scale changes to the way dependencies are handled that I believe address all the concerns there have been so far. I'd like people to re-review that area in particular, then I want to move this to final comment in a couple of days.

Boundarybreaker (approved):

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,388 @@
...
+            "quilt_rendering_api",
+            {
+                "id": "modmenu",
+                "environment": "client"
+            }
+        ],
+        "breaks": [
+            {
+                "id": "sodium",
+                "versions": "*",
```
joaoh1, line 274: Shouldn't this be `version` instead of `versions`?

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,388 @@
...
+                "reason": "Sodium does not implement the Quilt Rendering API."
+            },
+            {
+                "id": "some_random_library",
+                "versions": [
```
joaoh1, line 279: Uuh, little problem: there's nothing that defines a `versions` field that accepts an array of id-less dependency objects and/or version specifiers in this specification. Perhaps something about them should be written

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,388 @@
...
+| String | True     |
+
+A unique identifier for the organization behind or developers of the mod. See Maven's [guide to naming conventions](https://maven.apache.org/guides/mini/guide-naming-conventions.html).
+
+### The `id` field
+| Type   | Required |
+|--------|----------|
+| String | True     |
+
+A unique identifier or the mod or library defined by this file, matching the `^[a-z][a-z0-9-_]{1,63}$` regular expression. Best practice is that mod ID's are in snake_case.
```
joaoh1, line 61: I was going to complain about this but I end forgetting:
```suggestion
A unique identifier for the mod or library defined by this file, matching the `^[a-z][a-z0-9-_]{1,63}$` regular expression. Best practice is that mod ID's are in snake_case.
```

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,388 @@
...
+| String | True     |
+
+Must conform to the [Semantic Versioning 2.0.0 specification](https://semver.org/). In a development environment, the value `${version}` can be used as a placeholder by quilt-gradle to be replaced on building the resulting JAR.
+
+### The `entrypoints` field
+| Type   | Required |
+|--------|----------|
+| Object | False    |
+
+A collection of `key: value` pairs, where each key is the type of the entrypoints specified and each values is either a single entrypoint or an array of entrypoints. An entrypoint is an object with the following keys:
```
i509VCB, line 82: Inner details of what an entrypoint entry can look like probably needs further documentation.

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,419 @@
...
+| Array  | False    |
+
+An array of [dependency object](#dependency-objects)s. Defines mods that this mod either breaks or is broken by.
+
+### The `repositories` field
+| Type   | Required |
+|--------|----------|
+| Array  | False    |
+
+A list of repository url strings where dependencies can be looked for in addition to Quilt's central maven repository.
```
ramidzkh, line 165: Does this only support Maven?

Haven-King: Yes. What else would it support?

ramidzkh: It should clarify it then

famous1622: I feel like it's relatively self-explainatory

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,419 @@
...
+        }
+    },
+    "mixins": [
+        "modid.mixins.json"
+    ]
+}
+```
+
+## Placeholders
+Strings with defined formats such as [the version field](#the-version-field) can instead supply a string matching the regex `^\$\{[a-zA-Z_$][a-zA-Z\d_$]*\}$` in a development environment to be replaced on build. The quilt-gradle plugin defines usage of the value `${version}` as a placeholder for the version declared in gradle.properties.
```
lenrik1589, line 340: Can't `^\$\{[a-zA-Z_$][a-zA-Z\d_$]*\}$` be replaced with `^\$\{[a-zA-Z_$]+\}$` or the `+` not a standard regex?

i509VCB: I think this refers to a placeholder that gradle can replace at compliation time such as `{$version}`

Kroppeb: Those aren't the same character sets, although I didn't know you could use `\d` inside `[]`. I usually just use `0-9` like a normal person.

lenrik1589: > aren't the same character sets
Ah, sorry, my bad pretty hard to compare text on the phone, when it's a line break (because of scale)

loader/0002-quilt.mod.json.md
Chocohead: How so? You might be using an Ivy repo instead (as is easier to do if you use Github as a host)

loader/0002-quilt.mod.json.md
wagyourtail: It can be simplified to `^\$\{[a-zA-Z_$][\w$]*\}$` tho

loader/0002-quilt.mod.json.md
Haven-King: Maven is mentioned many times throughout the document, and even later in this same sentence. I can add some clarification, but it doesn't really seem necessary.

Haven-King:
Moving this to final comment. Final comment will conclude at 12PM UTC on May 12th.

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,419 @@
...
+                "Haven King": "Developer"
+            },
+            "contact": {
+                "homepage": "https://quiltmc.org/"
+            },
+            "license": "CC0-1.0",
+            "icon": "assets/modid/icon.png"
+        }
+    },
+    "mixins": [
```
joaoh1, line 333: The spec names that field as `mixin` while here it's `mixins`, is it supposed to be `mixin` or `mixins`?

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,419 @@
...
+            {
+                "id": "sodium",
+                "versions": "*",
+                "reason": "Sodium does not implement the Quilt Rendering API."
+            },
+            {
+                "id": "some_random_library",
+                "versions": [
+                    "1.23.456", // A reason is not required
+                    {
```
joaoh1, line 309: There's still no definition in the spec for those objects inside the version specifier array

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,423 @@
...
+        "group": "org.quiltmc",
+        "id": "example_mod",
+        "version": "1.0.0",
+        "entrypoints": {
+            "main": [
+                "org.quiltmc.example_mod.impl.ExampleMod",
+                "org.quiltmc.example_mod.impl.ExampleModNetworking"
+            ],
+            // Since we only have a single client endpoint, no array is needed.
+            "client": "org.quiltmc.example_mod.impl.client.ExampleModClient",
```
joaoh1, line 289: There's a trailing comma here:
```suggestion
            "client": "org.quiltmc.example_mod.impl.client.ExampleModClient"
```
I think this is the last nitpick I have for now

OroArmor: Building on that, `example_mod` for a package name isn't proper java naming. 
```suggestion
                "org.quiltmc.examplemod.impl.ExampleMod",
                "org.quiltmc.examplemod.impl.ExampleModNetworking"
            ],
            // Since we only have a single client endpoint, no array is needed.
            "client": "org.quiltmc.examplemod.impl.client.ExampleModClient",
```

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,423 @@
...
+            "quilt_rendering_api",
+            {
+                "id": "modmenu",
+                "environment": "client"
+            }
+        ],
+        "breaks": [
+            {
+                "id": "sodium",
+                "reason": "Sodium does not implement the Quilt Rendering API.",
```
joaoh1, line 302: Oh wait, I just noticed that there's no definition for the `reason` field in the spec, that should be rectified

duplexsystem:
Instead of hard coding [mixin] and [access_widener] wouldn't it be preferable to have something like [external_config] that takes an object like 
```json
{
    "mixin": "modid.mixins.json",
    "access_widener" : "modid.accesswidener",
    "custom_value" : "custom_file"
}
```
It would allow the ability to point to any external (from quilt.mod.json) config files. It could be used, for example, for a mod to get a list of mixins to disable, among other things.

Haven-King:
> Instead of hard coding [mixin] and [access_widener] wouldn't it be preferable to have something like [external_config] that takes an object like ...
Mods and loader plugins can define their own top-level custom values, as defined [here](https://github.com/QuiltMC/rfcs/blob/quilt.mod.json/loader/0002-quilt.mod.json.md#custom-elements). Mixin and access widener follow this convention.

joaoh1 (approved):
I think that being able to write a JSON Schema by only looking at the spec and having the example JSON being considered valid by it is pretty much a sign of success, and well, that has been achieved. Spec approved

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,430 @@
...
+| String | True     |
+
+A unique identifier for the mod or library defined by this file, matching the `^[a-z][a-z0-9-_]{1,63}$` regular expression. Best practice is that mod ID's are in snake_case.
+
+### The `provides` field
+| Type         | Required |
+|--------------|----------|
+| Array/String | False     |
+
+Either a single identifier or an array of identifiers that this mod also contains. These could be old or alternative mod ids. They should follow [the convention established for mod ids](#the-mod_id-field).
```
TheGlitch76, line 68: So one problem I've had when doing Quilt Loader is that our versions have diverged but our API is only compatible with Fabric Loader's API at a lower version. In the future, maybe the opposite happens and Fabric overtakes us, and then we have even more issues.

I think it should be possible to specify that you are providing a specific version of a dependency.

Haven-King: I've changed the `provides` field to use a dependency object instead of simple mod ids so that developers can provide various versions under different circumstances. I also added to the example to show what this might look like.

idijt (approved):

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,437 @@
...
+            },
+            "flamingo"
+        ],
+        "version": "1.0.0",
+        "entrypoints": {
+            "main": [
+                "org.quiltmc.examplemod.impl.ExampleMod",
+                "org.quiltmc.examplemod.impl.ExampleModNetworking"
+            ],
+            // Since we only have a single client endpoint, no array is needed.
```
williambl, line 295: Is this comment supposed to be here, since the json5-> json switch?

Haven-King: It's in the example to provide information that's relevant to the RFC. Mods will be expected to omit comments unless they are using a dev tool to remap JSON5 to JSON.

williambl: Cool, just checking since it was the only one and made the JSON invalid so seemed a little out of place

williambl (approved):

Redblueflame (approved):

loader/0002-quilt.mod.json.md
```diff
@@ -0,0 +1,437 @@
...
+
+## Dependency Objects
+A dependency object defines what mods/plugins a given mod depends on or breaks. It can be represented as either an object containing at least [the `id` field](#the-id-field-1), a string mod identifier in the form of either `mavenGroup:modId` or `modId`, or an array of dependency objects. If an array of dependency objects is provided, the dependency matches if it matches ANY of the dependency objects.
+
+### The `id` field
+| Type   | Required |
+|--------|----------|
+| String | True     |
+
+A mod identifier in the form of either `mavenGroup:modId` or `modId`.
```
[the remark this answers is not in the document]
Haven-King: No, duplicate mods should be determined based on mod ID alone.

loader/0002-quilt.mod.json.md
Haven-King: Each top-level element in the mod.json field should belong to a specific mod, loader plugin, or game provider. `mixin` and `access_widener` would theoretically be loader plugins, while `minecraft` is a game provider. A top-level `metadata` field would mean there would be a mod/plugin/game named `metadata`.

To summarize, top-level fields do not indicate purpose, but rather ownership.

loader/0002-quilt.mod.json.md
Haven-King: I believe the current format is somewhat clearer to those not familiar with the intricacies of regular expressions.

PR #9, opened 2021-04-19
RFC 9: QSL Structure

TheGlitch76:
[Rendered View](https://github.com/QuiltMC/rfcs/blob/qsl-structure/rfc/0009-qsl-structure.md)

### DIFF
--- a/rfc/0009-qsl-structure.md
+++ b/rfc/0009-qsl-structure.md
@@ -51,7 +51,7 @@ Modules follow the following general rules:
 To modders, most mods would not explicitly depend on libraries or their modules, but Loom would automatically determine the modules to depend on by scanning at compile time, and those modules will be downloaded at runtime by Loader. However, they will have the choice to manually depend on individual modules. There are no plans to provide fat jars for libraries or QSL as a whole at this time, besides as a temporary solution until Loader Plugins are implemented.
 ### Top-level Structure
 Keep in mind that the modules listed here are the names of the current Fabric API modules, and the final internal structure of QSL **will not** be the same as Fabric API.
-- `core` - All modules here are implicitly depended on by mods by default, even if they do not use any code from them. Mods can choose to opt-out.
+- `core` - All modules here are implicitly depended on by mods by default, even if they do not use any code from them. Mods can choose to opt-out. This library may be provided by the installer by default, in order to allow a handshake API to always be present.
     - Lifecycle Events
     - Networking API (maybe with relevant parts split into other modules)
         - Registry Sync


rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,134 @@
...
+    	- Content Registries (Flammable Block Registry)
+	- `items`
+    	- Item API
+    	- Content Registries (Fuel Registry, Composting Registry)
+    	- Tool Attribute API
+    	- (future) Item Extensions (think IForgeItem)
+	- `entities`
+    	- Object Builder API (EntityType Builder, Trades, Villager Professions and Types)
+    	- Entity Events
+    - `other`
```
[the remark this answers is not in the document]
Haven-King: `gui`, probably.

Haven-King:
> Let's say we have a tooltip component API (1.17 only), in which library would it go in?

`gui`, probably.

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,134 @@
...
+        - (future) Proper Handshake API
+    - Crash Reports
+    - Resource Loader
+- `common` - "Dumping ground" of shared code needed across modules. This is a little more complex, and uses github's code owners to define ownership of individual modules instead of the library as a whole. Modules should only be here if 100% needed, and if enough similar code goes here it can be split to its own library.
+	- None currently
+- `transfer`
+    - API Lookup API
+    - fluid api
+    - item transfer api
+- `config_and_admin` // better name needed
```
quat1024, line 69: Commands and gamerules don't really strike me as something related to `config` - i'm thinking `command`, maybe (in my mind permissions are typically used to restrict access to commands bukkit-style, but i dunno what the permissions api workgroup is up to)

quat1024:
[Rendered markdown](https://github.com/TheGlitch76/rfcs/blob/qsl-structure/rfc/0009-qsl-structure.md) btw

Big fan of the two-tier structure!!

some knee-jerk reactions:

* `world` is kind of big, especially if each of its submodules is to be managed by the same team.
* Where does a config api fit? `config_and_admin` has `config` in the name but seems to deal more with "ingame" things. Maybe it fits in `core`?

i509VCB:
I wouldn't say the commands, gamerules or permissions stuff would belong in core.

For sake of not making core huge (I want to be very selective about what makes the cut for core, config would probably be seperate).

i509VCB:
Keybindings do apply only in a world, but it doesn't feel very keybind like tbf. I guess it's the cost of consolidating under more broad categories.

quat1024:
> I wouldn't say the commands, gamerules or permissions stuff would belong in core.

Oh i absolutely agree lol, by "config api" i meant something like "loading and parsing a config file"

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,134 @@
...
+	- Textures
+	- Object Builder API (FabricModelPredicateProviderRegistry)
+- `data` - Tools for working with, generating, or consuming data (e.g. datapacks, advancements, recipes, tags)
+    - (future) Recipe API
+    - (future) Datagen APIs
+    - Rendering Data Attachment
+    - Tag Extensions
+    - Loot Tables
+    - Object Builder API (Criterion)
+- `world` // Needs better name to avoid confusion with Worldgen
```
LambdAurora, line 92:
```suggestion
- `ingame`
```

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,134 @@
...
+    	- (future) Item Extensions (think IForgeItem)
+	- `entity`
+    	    	- Object Builder API (EntityType Builder, Trades, Villager Professions and Types)
+    	    	- Entity Events
+    - `other`
+    	- Particles
+    	- Object Builder API (Point of interest)
+    	- Interaction Events
+        - Could probably use refactoring, but as it stands this absolutely deals with events happening in the world
+    	- Key Binding API
```
LambdAurora, line 111: Key Binding API should not be in the "ingame" library.

After quickly checking the code key bindings can work in screens that are not in-game.

In Keyboard we can see the condition `if (this.client.currentScreen == null || this.client.currentScreen.passEvents)` before changing states of key bindings.

LambdAurora (changes requested):

l-Luna:
why require modules to "only depend on modules from `core`, `common`, or within their own library"? it seems to me that the easiest way to make indigo easy to exclude would be to have it as a seperate module that depends on the rendering module. otherwise great

TheGlitch76:
>  it seems to me that the easiest way to make indigo easy to exclude would be to have it as a seperate module that depends on the rendering module.

Yes, for mods that need to exclude Indigo I was thinking something like:
```
implementation qsl(...) {
    exclude module 'rendering.indigo'
}
```

> why require modules to "only depend on modules from core, common, or within their own library"?

Consider a "minimal" mod like Sodium, which is avoiding pulling in game-changing behavior because it is used by the Minecraft Speedrunning community. (This is all hypothetical)
In its buildscript, it defines:
```
implementation qslLibrary("rendering") {
    autoDepend() // instead of always puling in all the rendering modules, it only depends on ones that it uses
    exclude module 'indigo'
}```
This works fine for Sodium, perhaps for months. Later, though, Sodium calls a single line from a rendering module that depends on, say, something from block extensions.
Now, with no change to their buildscript, they are now pulling in unwanted transformations and breaking the Minecraft Speedrunning rules! Additionally, this wouldn't be immediately obvious from Sodium's `quilt.mod.json`, since you would only see it depending on rendering modules, not block extensions.

In short, we want to prevent unexpected behavior when you think you're depending on something explicitly.

File: rfc/0009-qsl-structure.md (hunk 2 above)
Comment (redcreeper14385) [reply rfc/0009-qsl-structure.md:69]: I think `server` would be better, as they are all server-related (integrated or otherwise). They could even go under `data` I reckon.

File: rfc/0009-qsl-structure.md [hunk 5]
```diff
@@ -0,0 +1,134 @@
...
+	- `item`
+    	- Item API
+    	- Content Registries (Fuel Registry, Composting Registry)
+    	- Tool Attribute API
+    	- (future) Item Extensions (think IForgeItem)
+	- `entity`
+    	    	- Object Builder API (EntityType Builder, Trades, Villager Professions and Types)
+    	    	- Entity Events
+    - `other`
+    	- Particles
```
redcreeper14385, line 107: This can go under `rendering`.

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,134 @@
...
+    	- (future) Block Extensions (think IForgeBlock)
+    	- Content Registries (Flammable Block Registry)
+	- `item`
+    	- Item API
+    	- Content Registries (Fuel Registry, Composting Registry)
+    	- Tool Attribute API
+    	- (future) Item Extensions (think IForgeItem)
+	- `entity`
+    	    	- Object Builder API (EntityType Builder, Trades, Villager Professions and Types)
+    	    	- Entity Events
```
redcreeper14385, line 105: This could all go into a `registry` section.

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,134 @@
...
+    - `other`
```
redcreeper14385, line 106: Is this part of `world` or not? It's not clear.

redcreeper14385 (commented):
A few comments here and there, LGTM otherwise.

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,134 @@
...
+	- `block` // these groups are 100% internal; they are in the same repo and all managed by the `world` team
+		- BlockEntity Networking (merged in)
+    	- Object Builder API (FabricBlockSettings, FabricBlockEntityBuilder and FabricMaterialBuilder)
+    	- (future) Block Extensions (think IForgeBlock)
+    	- Content Registries (Flammable Block Registry)
+	- `item`
+    	- Item API
+    	- Content Registries (Fuel Registry, Composting Registry)
+    	- Tool Attribute API
+    	- (future) Item Extensions (think IForgeItem)
```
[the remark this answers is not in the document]
TheGlitch76: The point of this comment is that there isn't a current Fabric API module with that name.

rfc/0009-qsl-structure.md
TheGlitch76: Particles are [server-side constructs](https://github.com/FabricMC/fabric/blob/1.17/fabric-particles-v1/src/main/java/net/fabricmc/fabric/api/particle/v1/FabricParticleTypes.java) so it doesn't make sense to put them in a module dealing with client-only features?
I will update this RFC to note that in the scope of the rendering module.

TheGlitch76:
> world is kind of big, especially if each of its submodules is to be managed by the same team.

I agree, recommendations on splitting up the `world` library would be appreciated.

I don't think there's a purpose in making `blocks` and `items` their own libraries, because they would probably be co-dependent.
Perhaps there could be a split between `extensions` (namely block, item, entity) and then everything else?

rfc/0009-qsl-structure.md
redcreeper14385: OK 👍

i509VCB (commented):
The structure here still feels confused is my issue atm.

rfc/0009-qsl-structure.md
TheGlitch76: I have chosen `command` as a name for now, if anyone has a *better* name ("command isn't descriptive" is not productive) please let it be known.

rfc/0009-qsl-structure.md
TheGlitch76: I don't feel `registry` is a good fit for this, because it also involves extensions and behavior changes (like Item API)

rfc/0009-qsl-structure.md
TheGlitch76: No longer relevant after flattening the `world` library

TheGlitch76:
I believe I have covered all concerns now.

OroArmor (approved):

rfc/0009-qsl-structure.md
```diff
@@ -60,13 +60,11 @@ Keep in mind that the modules listed here are the names of the current Fabric AP
...
-	- None currently
 - `transfer`
     - API Lookup API
     - fluid api
     - item transfer api
-- `config_and_admin` // better name needed
+- `command` // This really is anything server-side only related to administration, but "command" is a nice word to describe the primary use-case.
     - Command API
     - Game Rule API
     - (future) Permissions API
```
i509VCB, line 70: I feel permissions will probably exceed the scope of commands in QSL

rfc/0009-qsl-structure.md
```diff
@@ -60,13 +60,11 @@ Keep in mind that the modules listed here are the names of the current Fabric AP
...
-- `common` - "Dumping ground" of shared code needed across modules. This is a little more complex, and uses github's code owners to define ownership of individual modules instead of the library as a whole. Modules should only be here if 100% needed, and if enough similar code goes here it can be split to its own library.
-	- None currently
 - `transfer`
     - API Lookup API
     - fluid api
     - item transfer api
-- `config_and_admin` // better name needed
+- `command` // This really is anything server-side only related to administration, but "command" is a nice word to describe the primary use-case.
     - Command API
     - Game Rule API
```
i509VCB, line 69: I'm not sure this is appropriate still.

Given there is a game rule screen and it really has a world scope.

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,133 @@
...
+- Semantic Versioning. Hashes are *not* used to identify modules. 
+- Can only depend on modules from `core` or within their own library.
+
+To modders, most mods would not explicitly depend on libraries or their modules, but Loom would automatically determine the modules to depend on by scanning at compile time, and those modules will be downloaded at runtime by Loader.
+### Top-level Structure
+Keep in mind that the modules listed here are the names of the current Fabric API modules, and the final internal structure of QSL **will not** be the same as Fabric API.
+- `core` - All modules here are implicitly depended on by mods by default, even if they do not use any code from them. Mods can choose to opt-out from this, but if they choose to use the fat QSL dependency they are responsible for ensuring these modules don't get transitively pulled in.
+    - Lifecycle Events
+    	- Here because of the Networking API
+    	- Some parts might be split into `entities` and `world`
```
i509VCB, line 57: World is gone fyi

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,133 @@
...
+		- This may need a better name in a spot?
+	- Rendering
+	- Renderer Registries
+	- Rendering Fluids
+	- Textures
+	- Object Builder API (FabricModelPredicateProviderRegistry)
+- `data` - Tools for working with, generating, or consuming data (e.g. datapacks, advancements, recipes, tags)
+    - (future) Recipe API
+    - (future) Datagen APIs
+    - Rendering Data Attachment
```
i509VCB, line 86: Rendering data attachment is not exactly related to generation or modification or consumption of data.

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,133 @@
...
+- Artifact id of `mod-id`
+- Semantic Versioning. Hashes are *not* used to identify modules. 
+- Can only depend on modules from `core` or within their own library.
+
+To modders, most mods would not explicitly depend on libraries or their modules, but Loom would automatically determine the modules to depend on by scanning at compile time, and those modules will be downloaded at runtime by Loader.
+### Top-level Structure
+Keep in mind that the modules listed here are the names of the current Fabric API modules, and the final internal structure of QSL **will not** be the same as Fabric API.
+- `core` - All modules here are implicitly depended on by mods by default, even if they do not use any code from them. Mods can choose to opt-out from this, but if they choose to use the fat QSL dependency they are responsible for ensuring these modules don't get transitively pulled in.
+    - Lifecycle Events
+    	- Here because of the Networking API
```
i509VCB, line 56: Useful enough to have in core to where I don't think that reasoning is needed.

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,133 @@
...
+    	- Some parts might be split into `entities` and `world`
+    - Networking API (maybe with relevant parts split into other modules)
+        - Registry Sync (merged in)
+        - (future) Proper Handshake API
+    - Crash Reports
+    - Resource Loader
+- `transfer`
+    - API Lookup API
```
i509VCB, line 64: Obviously some bikeshed, but I'd like to shorten this to `lookup api`.

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,133 @@
...
+    - Content Registries (Fuel Registry, Composting Registry)
+    - Tool Attribute API
+    - (future) Item Extensions (think IForgeItem)
+    - Item Groups
+- `entity`
+    - Object Builder API (EntityType Builder, Trades, Villager Professions and Types)
+    - Entity Events
+- `content_other`
+    	- Particles
+    	- Object Builder API (Point of interest)
```
i509VCB, line 106: You would need to implicitly depend on content_other to also register villager types from entity api fyi. Though it is more appropriate here.

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,133 @@
...
+    	- Interaction Events
```
i509VCB, line 107: I'm actually thinking entity for this, it is in scope of a player interacting with the world. And a player is an entity.

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,133 @@
...
+        	- Could probably use refactoring, but as it stands this absolutely deals with events happening in the world
+- `worldgen`
+    - Dimension API
+    - Biome API
+    - Structure API
+- `gui`
+    - Screen API
+        - Despite the name, all this module really does is add hooks for modifying prexisting screens
+    - ScreenHandler API
+        - this could also go in `world`
```
i509VCB, line 117: I'd probably say gui more likely that world for this. This api specifically addresses registration of the handled screen and the screen handler type.

i509VCB (commented):
Those are my comments about putting the puzzle pieces in place, I am going to make another set of comments about the question of this RFC in general.

rfc/0009-qsl-structure.md
Technici4n: You don't want a Quilt API API-API API Lookup API?

TheGlitch76: This isn't the place to bikeshed internal module names.

rfc/0009-qsl-structure.md
TheGlitch76: hm, do you have a better idea for the library name then?

i509VCB:
Okay so I forgot to respond, so my issue with the RFC is similar to the community team one. The issue I see with ironing the structure of QSL here with specific modules existing makes a little confusing, do new modules need an RFC, how about collapsing modules.

What I think we could do instead is require a document to be maintained regarding the structure of qsl and how to setup of modules that is maintained by the team specifically.

TheGlitch76:
> do new modules need an RFC, how about collapsing modules.

From the very first line of the Explanation:
> Libraries are included along with the names of the FAPI modules that would be under its scope. The actual structure of the modules in each library is up to the team, and it is not a goal to define how they must structure their modules.

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,129 @@
...
+Modules follow the following general rules:
+- Maven group of `org.quiltmc.qsl.$library`
+- Artifact id of `mod-id`
+- Semantic Versioning. Hashes are *not* used to identify modules. 
+- Can only depend on modules from `core` or within their own library.
+
+To modders, most mods would not explicitly depend on libraries or their modules, but Loom would automatically determine the modules to depend on by scanning at compile time, and those modules will be downloaded at runtime by Loader.
+### Top-level Structure
+Keep in mind that the modules listed here are the names of the current Fabric API modules, and the final internal structure of QSL **will not** be the same as Fabric API.
+- `core` - All modules here are implicitly depended on by mods by default, even if they do not use any code from them. Mods can choose to opt-out.
```
gdude2002, line 54: > Mods can choose to opt-out.

This seems unworkable to me, from a community standpoint. From the Quilt community rules (5, specifically)

> Projects must not be malicious - which includes (but is not limited to) malware, mods that include backdoors, mods that modify Quilt in an attempt to hide themselves from servers they connect to, or mods that harass specific people or groups of people (for example, by automatically kicking them from servers with the mod installed)

While this RFC doesn't mess with the wording of the rule, it does mess with the spirit - we've been working under the assumption that whatever modlist/modid handshake logic would always be present - being able to opt out by simply saying "No, thanks" is antithetical to the goal of giving servers control over the mods that their users are making use of.

Whether this means it needs to be factored out into a separate module and have that be required, or whether there's some other solution - that's fine. But providing machinery directly within Quilt to entirely bypass this technical measure will make it much harder for the community team to do their jobs. While the best approach from that aspect may simply be "also disallow mods that opt-out", it still raises the burden of moderation because it's something we have to actually find a way to check directly.

I'm sure there's a more workable approach than this.

gdude2002 (changes requested):

rfc/0009-qsl-structure.md
i509VCB: I'm going to say the concerns about rule 5 are probably not an issue here regarding the technical process. The ability to opt out of using qsl entirely does not appear to me as circumvention or malicious and is rather implementation detail of a mod.

A mod does not need to opt-in to being included in any metadata handshake, that operates off loader's api.

gdude2002: Overall I agree with Glitch's approach - there are a lot of concerns to think of here that aren't necessarily related to technology; Glitch's approach seems like the best middle ground to me.

Modularity is a good thing, yes - I've been preaching it myself actually. The idea of something being used to gate users not being included by default is a tricky one, IMO, since most users only use optimisation mods that are unlikely to require QSL, meaning their experience would be broken by servers that require the packet - so including a checkbox or something in the installer that allows them to opt in when no mods use QSL seems like a good solution.

Of course, that could have its own technical problems - like putting it on the classpath and breaking mods that require other verisons. It's a tricky one

TheGlitch76: > Of course, that could have its own technical problems - like putting it on the classpath and breaking mods that require other verisons. It's a tricky one

I don't see this as much of a concern--we can tell Loader to download it the same way it would if some mod depended on it

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,129 @@
...
+    - Networking API (maybe with relevant parts split into other modules)
+        - Registry Sync
+        - (future) Proper Handshake API
+    - Crash Reports
+    - Resource Loader
+- `transfer`
+    - API Lookup API
+    - fluid api
+    - item transfer api
+- `command
```
LambdAurora, line 65:
```suggestion
- `command`
```

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,129 @@
...
+- Avoid making users download and run code they don't need
+## Explanation
+This is an overview of the top-level structure of QSL. Libraries are included, with the names of the FAPI modules that would be under its scope. The actual structure of the modules in each library is up to the team, and it is not a goal to define how they must structure their modules.
+### Libraries
+Each library would be its own folder in the QSL monorepo, with a dedicated team responsible for triage, review, and accepting/denying pull requests. In contrast to Fabric or Forge's system of only one person signing off for the entire standard library, no authority higher than the people on the team would be required to sign off on the PR before it is merged. More clarification on team structure is expected to come in a future RFC.
+
+Libraries only exist for internal structure--there is no fat jar for, say, all `worldgen` modules, neither is there a fat jar for all of QSL. This also means that QSL and its libraries have no versioning.
+### Modules
+Modules follow the following general rules:
+- Maven group of `org.quiltmc.qsl.$library`
```
LambdAurora, line 46: I have to say that might end up a bit confusing for anyone having to deal with a maven group.

LambdAurora (changes requested):
Currently I see nothing about the state of libraries and modules exposure to maven repositories.

Can a modder still depend on a module alone or only libraries are exposed?

Would be cool if there was a word on that to make it more clear.

rfc/0009-qsl-structure.md
TheGlitch76: Since we've found a workable solution that's out of scope of QSL, I'm going to consider this issue solved for now.

TheGlitch76:
I am proposing a Final Comment Period for this RFC. It will begin once both i509 and LambdAurora have approved this pull request (the full membership of the QSL team at the time of writing), and last the standard ten days.

LambdAurora (approved):

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,129 @@
...
+- Find a balance between Forge's confusingly monolithic and Fabric's confusingly modular structures
+- Have teams in charge of different parts of QSL's code
+- Avoid making users download and run code they don't need
+## Explanation
+This is an overview of the top-level structure of QSL. Libraries are included, with the names of the FAPI modules that would be under its scope. The actual structure of the modules in each library is up to the team, and it is not a goal to define how they must structure their modules.
+### Libraries
+Each library would be its own folder in the QSL monorepo, with a dedicated team responsible for triage, review, and accepting/denying pull requests. In contrast to Fabric or Forge's system of only one person signing off for the entire standard library, no authority higher than the people on the team would be required to sign off on the PR before it is merged. More clarification on team structure is expected to come in a future RFC.
+
+Libraries only exist for internal structure--there is no fat jar for, say, all `worldgen` modules, neither is there a fat jar for all of QSL. This also means that QSL and its libraries have no versioning.
```
yitzy299, line 43: Without a fat jar for QSL before auto dependency downloading, running mods which depend on QSL would be hell. And even then, if the auto downloading fails that's still a lot of jars to manually install

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,129 @@
...
+### Modules
+Modules follow the following general rules:
+- Maven group of `org.quiltmc.qsl.$library`
+- Artifact id of `mod-id`
+- Semantic Versioning. Hashes are *not* used to identify modules. 
+- Can only depend on modules from `core` or within their own library.
+
+To modders, most mods would not explicitly depend on libraries or their modules, but Loom would automatically determine the modules to depend on by scanning at compile time, and those modules will be downloaded at runtime by Loader. However, they will have the choice to manually depend on individual modules. There are no plans to provide fat jars for libraries or QSL as a whole at this time, besides as a temporary solution until Loader Plugins are implemented.
+### Top-level Structure
+Keep in mind that the modules listed here are the names of the current Fabric API modules, and the final internal structure of QSL **will not** be the same as Fabric API.
```
yitzy299, line 53: A language module would be useful, for Quilt-language-kotlin and the like

LambdAurora: imo this should be separate from QSL.

Languages adapters should stay their own thing.

TheGlitch76: Future additions is mostly out of scope for this RFC, so I'm going to resolve this for now.

TheGlitch76:
I've changed up the wording on where we will go involving fat jars in QSL a bit.
The relevant section now reads:

> Ideally, there would be no common usage "fat" jars of libraries and modules because of Loader's automatic dependency downloading. We will probably have to provide fat jars until Loader is able to support dependency downloading, and discussion about stopping providing fat jars can happen there.

rfc/0009-qsl-structure.md
TheGlitch76: You're right--I've changed this section a bit to not strongly require either approach.

LePichu (approved):
I approve of these changes.

i509VCB (approved):
If we need to move features between libraries, we do that later before initial release.

famous1622 (approved):
LGTM

yitzy299 (approved):

LambdAurora (approved):

Platymemo (approved):

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,129 @@
...
+- `core` - All modules here are implicitly depended on by mods by default, even if they do not use any code from them. Mods can choose to opt-out. This library may be provided by the installer by default, in order to allow a handshake API to always be present.
+    - Lifecycle Events
+    - Networking API (maybe with relevant parts split into other modules)
+        - Registry Sync
+        - (future) Proper Handshake API
+    - Crash Reports
+    - Resource Loader
+- `transfer`
```
hYdos, line 61: maybe `fluid` should be above this. i dont think a general fluid api is all transfer

hYdos (changes requested):
e

rfc/0009-qsl-structure.md
Platymemo: I believe the fluid API mentioned here is specifically the fluid transfer API, but clarification would be nice. As for a fluid API for creating in-world fluids, I feel it may fit as a subsection within the `block` library.

grondag (approved):
Something to consider about Indigo is that the use cases are a little more elaborate than excluding it.  Mods should never depend on Indigo directly but instead depend on the Rendering API and QSL should try to ensure the "best available" implementation of that API is loaded. 

There are at least three common situations:

1) Nothing depends on or implements the Rendering API, in which case Indigo (and the rendering API) wouldn't be loaded. (That said - Indigo does have some bug fixes and minor improvements, so should it be loaded anyway?  Should this be configurable?)

2) A mod breaks /conflicts with Indigo and/or the rendering API.  Sodium is the most notable example here.  Not only should Indigo not be loaded, but any mod that depends on the Rendering API it implements cannot run with Sodium. (But see the exception below...)

3) An alternate implementation of the rendering API is present - Canvas is an example. Indigo should not load but the rendering API must and mods can depend on it.  (The renderer is a singleton, so it's a problem if more than one implementation tries to load, but it's not a a common problem yet.  As long as it's obvious when it happens that's probably good enough.)

An interesting exception to 2 is Indium, a mod that adds support for the rendering API to Sodium.  When Sodium is present by itself we should have the behavior of 2, but when Indium is added the outcome should be 3.

Leo40Git (approved):

rfc/0009-qsl-structure.md
```diff
@@ -0,0 +1,129 @@
+# QSL Structure
+## Summary
+This documents provides the basis on how to structure the Quilt Standard Library. It seeks to define how Fabric's current API will be split into several libraries (collections of modules) for the purposes of development. It does not define the structure of the modules within these libraries; that is left to the teams discretion when creating QSL, besides the guideline that modules should be generally larger than what is expected in Fabric API.
```
LeoCTH, line 3: `teams` -> `teams'`

LeoCTH (approved):

TheGlitch76:
Apologies for the delay; I was on vacation when the FCP criteria were met.
The Final Comment Period has begun and will end on Tuesday, June 22nd, at 00:00 UTC

williambl (approved):

PR #11, opened 2021-04-21
Amendment to RFC 7: Address negative vote fast-failure

sciwhiz12:
### Amend the negative greater than positive votes clause in "Interpreting the Vote"

There may be a situation where the following happens:

 * An election for a candidate has started, with 0 total votes.
 * An opposing voter casts their votes against the candidate, leaving the poll at 0 positive votes and 1 negative vote.
 * Due to this clause, the election _fails immediately_ because there are a greater amount of negative votes than positive votes (1 negative > 0 positive).

This allows (even accidentally) for a voter to singlehandedly fail an election, even if the candidate would have been passed by the whole voter pool.

To prevent this possibility, the condition is amended to fail the election if an absolute majority of the voters have voted against the candidate, instead of only a simple majority.

Originally, I intended to remove the clause entirely, but further discussion in the Community discord led me to change from removal to amendment; the clause was originally intended to allow an election to fail immediately if more than half of all the possible votes (an absolute majority). This amends the clause to follow the original spirit.

### DIFF
--- a/rfc/0007-community-team.md
+++ b/rfc/0007-community-team.md
@@ -233,7 +233,7 @@ The election has **failed** when one of the following is true:
 * A week has passed and less than 75% of the community team and administrative board has voted
 * A week has passed, at least 75% of the community team and administrative board has voted, and those that have
   voted have reached a consensus in opposition of the candidate - or a consensus cannot be reached
-* The candidate receives more negative votes than positive votes
+* An absolute majority of the community team and administrative board has voted against the candidate
 
 If a situation occurs that this section is unable to cover, it will be amended. In those situations, the election will
 fail and need to be restarted after the relevant changes have been made to address the issue.
--- a/rfc/0007-community-team.md
+++ b/rfc/0007-community-team.md
@@ -233,7 +233,7 @@ The election has **failed** when one of the following is true:
 * A week has passed and less than 75% of the community team and administrative board has voted
 * A week has passed, at least 75% of the community team and administrative board has voted, and those that have
   voted have reached a consensus in opposition of the candidate - or a consensus cannot be reached
-* An absolute majority of the community team and administrative board has voted against the candidate
+* The candidate recieves more negative votes than the total amount of positive votes and votes not yet submitted
 
 If a situation occurs that this section is unable to cover, it will be amended. In those situations, the election will
 fail and need to be restarted after the relevant changes have been made to address the issue.


Earthcomputer:
> An absolute majority of the community team and administrative board has voted against the candidate

What if a fair amount of people on such teams become inactive? I think it's better to set a time limit for voting, and take the count a certain number of hours/days after the vote was posted.

Scotsguy:
This is only one of the fail conditions, it already times out

Earthcomputer:
I am not talking about a maximum about of time for the vote to take place, I am talking about a minimum amount of time before the count can be taken.

Scotsguy:
That's only relevant if people change their mind, which I don't think you were getting at in your initial comment.

Scotsguy (approved):

Haven-King (approved):

Earthcomputer (changes requested):
Err... so what happens when 51% of the community members don't vote because they're inactive? No vote could ever possibly pass with this rule.

Haven-King:
> Err... so what happens when 51% of the community members don't vote because they're inactive? No vote could ever possibly pass with this rule.

This voting process is internal to staff members. It's not 51% of all community members, but rather, 51% of the community *team* members (plus admin board). Though really, this should be made more generalized, as votes are not specific to the community team.

Earthcomputer:
Yes I meant community team members

rfc/0007-community-team.md
```diff
@@ -233,7 +233,7 @@ The election has **failed** when one of the following is true:
 * A week has passed and less than 75% of the community team and administrative board has voted
 * A week has passed, at least 75% of the community team and administrative board has voted, and those that have
   voted have reached a consensus in opposition of the candidate - or a consensus cannot be reached
-* The candidate receives more negative votes than positive votes
+* An absolute majority of the community team and administrative board has voted against the candidate
```
Earthcomputer, line 236: I would change it to something like this:
```suggestion
* An absolute majority of the community team and administrative board who voted, has voted against the candidate, after 3 days of the vote being open.
```

Haven-King: Your change goes completely against the purpose of this PR, as that is a simple majority, and is actually covered by this portion of the existing RFC:

> A week has passed, at least 75% of the community team and administrative board has voted, and those that have
  voted have reached a consensus in opposition of the candidate

Essentially, this PR breaks down to:
- If 75% of the people who *actually voted* vote negatively, the vote fails (existing rule)
- *Additionally*, if 51% of everyone who *could* vote votes negatively, the vote fails

Earthcomputer: I am worried about a situation where if the project is already dying, this system will cause it to enter a spiral which kills it faster. If we ever reach that 25% threshold of community team and admin team members becoming inactive for some reason, it would be impossible for any election to pass and to elect any active team members.

Haven-King: If anything, this change makes that *less* likely to be a problem, as the inactive members cannot vote explicitly against the candidate. All this change does is increase the threshold for a failed vote from "If there are ever more negative voters than positive voters, it fails" to "If more than half of everyone who *could* vote *actively votes against the candidate*, the vote fails".

parzivail: Wouldn't this whole issue be covered by the fact that people deemed too inactive to be beneficial to the project can be democratically voted out of their position on the team, and/or someone more active can be added to the team?

Earthcomputer: Hmm, I think for clarity the "a vote passes when" section needs to be amended to reflect this. Because right now there is no bullet point under that section that says a vote could pass if less than 75% of the team is active.

sciwhiz12: In my opinion, it would be better to have an RFC which centralizes and consolidates voting processes. Such an RFC would tie in to RFC 7 and to [RFC 6: Governmental Structure](https://github.com/QuiltMC/rfcs/pull/6) (such as for votes for member removal or team creation).

This PR is currently scoped to amending the specific "negative votes greater than positive votes" clause, which is a more pressing matter if an election were to be held (because this can be triggered by one team member, vs the 25% membership count that would block a vote if they are inactive). 

While I am not against clarifying how a vote can pass if less than 75% of the team has voted and 25%+ are 'inactive', as I've said, it would best be resolved by an RFC which details the voting process (such as inactivity, abstaining, majorities required for different actions, uncovered situations, etc) so everything voting-related is linked in one central place.

gdude2002, line 236: I would reword this.

> The candidate recieves enough negative votes to outweigh the remaining count of possible positive votes

That's the intention of this point, and I think this wording captures the style I used in the rest of the RFC better.

gdude2002 (changes requested):

gdude2002:
In accordance with this PR, I've updated the [community tasks project board](https://github.com/orgs/QuiltMC/projects/2).

I'd appreciate a mention on Discord or some other notification before the community decides to handle a task on that board, as it allows us to work directly with you and advise you on an approach before you get started.

Not a big deal for this PR, but I thought it was worth mentioning.

rfc/0007-community-team.md
sciwhiz12: In my opinion, `the remaining count of possible positive votes` is a bit too hard to digest, and can be interpreted as "the count of votes that could possibly be positive, but aren't yet" (because `possible` has the connotation of 'able to be done, but not yet done').

May I suggest the alternative of:
> The candidate recieves more negative votes than the total amount of positive votes and votes not yet submitted

gdude2002: I guess that's fine too, yeah.

sciwhiz12: Done.

gdude2002 (approved):

rfc/0007-community-team.md
Kroppeb: Tbh I find this more confusing than the original.

gdude2002: How do you figure?

`negative > (positive + undecided)`

Kroppeb: Using a mathematical expression seems weird. 

I checked on Wikipedia where it mentions that "absolute majority" often means a majority of all those who can vote, and not just all those who have voted, but that this definition is not always used. 

But because of the fact it says "Absolute majority of the community team and administrative board", there shouldn't be any confusion possible.

gdude2002: I don't really think that's a necessary clarification, especially given that this is an internal process for the most part - it's important that the RFC accurately describes the process, but I don't think it necessarily needs to completely avoid every chance of needing to be clarified for some people.

We should avoid bikeshedding this.

PR #12, opened 2021-04-21
Amendment to RFC 9: Add experimental features.

Redblueflame:
[Rendered](https://github.com/Redblueflame/rfcs/blob/qsl-structure/rfc/0009-qsl-structure.md)

Everything needed for the initial RFC is here, but there will be more to explain in a separate RFC once this one is merged, and working groups are ready.

## Reasoning

Experimental features are really needed to prevent incrementing the major / minor version every time a change is made to the user facing APIs, and allow for staged testing before a general availability.

This system is used successfully in rust's standard library, and bring safety for developers, that are sure of the quality of each function called.

### DIFF
--- a/rfc/0009-qsl-structure.md
+++ b/rfc/0009-qsl-structure.md
@@ -121,10 +121,13 @@ Keep in mind that the modules listed here are the names of the current Fabric AP
         - this could also go in `world`
     - (potentially in the future) Screen Extensions
 ## Experimental features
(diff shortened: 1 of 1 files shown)


i509VCB:
Would probably be good to note that `@ApiStatus.Experimental` and `@Deprecated` should both be applied to apis so IDE notification works out.

Redblueflame:
> Would probably be good to note that `@ApiStatus.Experimental` and `@Deprecated` should both be applied to apis so IDE notification works out.

True, but isn't it a little bit too close to the implementation for the scope of this amendment? I'm currently working on a separate RFC for the specifics of the experimental status, in addition to all the other lifetimes.

TheGlitch76:
Yeah, let's keep the annotation stuff to RFC 13, which formally defines the implementation of experimental features.

TheGlitch76 (approved):

PR #15, opened 2021-04-23
RFC 15: Data Generation

Boundarybreaker:
Data generation has been a major challenge since 1.13, and I've been musing on ways to improve it. **Still very not final**; suggestions for alternative static datagen methods are more than welcome.

### DIFF
--- a/rfc/0015-data-generation.md
+++ b/rfc/0015-data-generation.md
@@ -1,0 +1,140 @@
+## Summary
+
+This document defines a scheme for supporting both static and dynamic data
+generation, so that modders can spend less time writing JSONs while still
+keeping the advantages of statically-defined data when dynamic generation
+is unnecessary.
+
+
+## Motivation
+
+Since the beginning of The Datapack Era, modders have consistently expressed
+frustration with the amount of JSON files that need to be written. A standard
+block will require:
+
+- A blockstate JSON to define what models to use
+- Separate model JSONs for each permutation of visual change
+- A separate model JSON for the item model
+- A separate JSON for each recipe to obtain the block
+- A JSON for each tag the block belongs to
+- A JSON for each tag the *item form* of the block belongs to
+- A JSON for unlocking the block's recipe through an advancement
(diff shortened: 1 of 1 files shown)


i509VCB:
One thing this does need to consider is IDE support since you need to either track changes or tell the ide about the generated assets in order to place assets on classpath.

Boundarybreaker:
...oh, right, IDE support, delicious, forgot about that

Boundarybreaker:
Another proposal is to use a specific source set for compile-time data gen. Gonna rework the RFC for that later.

Boundarybreaker:
I've written up a simple description of the new system, ready for further feedback.

rfc/0015-data-generation.md
```diff
@@ -0,0 +1,132 @@
...
+`quilt_generated` folder in the base game directory. The state object will
+be written as well, and compared to whenever a resource reload is triggered.
+If the current state and the saved state are different, the resources will be
+regenerated, allowing for reliable caching.
+
+## Drawbacks
+
+- Getting multiple source sets, especially arbitrary source sets, to play nice
+with IDEs can be challenging. This would require work in Quilt Gradle to ensure
+that source sets are executed on any build, whether it's to test or to release.
```
LeoCTH, line 84: I had made several (3 in total) data generators for MC modding, and I can whole-heartedly tell you that IntelliJ IDEA and Gradle just do not work with multiple source sets/modules and annotation processors; I often need to regenerate IDEA modules since the mapping from Gradle subprojects/source sets often get messed up, my annotation processor sometimes refuses to work in the test mod, and the main reason I stopped developing that data generator is that Mixins just don't apply *at all*, and so Fabric API crashes spetacularly.
Maybe this is a complex issue that is related to how IDEA, its Gradle plugin and Fabric/Quilt Loom exactly works, but to my experience it's not really pleasant.

Boundarybreaker:
I'm expecting this RFC is gonna hibernate for a bit until we resolve QSL structure and the like

PR #16, opened 2021-04-23
RFC 16: Specification Repository

TheGlitch76:
Well, the meeting that should have been an email has become a meeting again.

[Rendered View](https://github.com/theglitch76-forks/rfcs/blob/specification-repo/rfc/0016-specification-repository.md)

### DIFF
--- a/rfc/0016-specification-repository.md
+++ b/rfc/0016-specification-repository.md
@@ -30,6 +30,9 @@ Each specification only contains the actionable parts of the RFCs involved (usua
 
 - Creation of a specification for technical definitions is at the author's discretion, but may be required in cases where major parts of the document are expected to be revised, or if an amendment PR to the RFC repository is deemed as "too big" and needing its own RFC.
 
+## Other details
+How general policy is split into specifications is not meant to be 1 rfc -> 1 file; maintainers should go for clarity and ease of reference, with refactors where needed.
+
 # Drawbacks
 
 - This change, if accepted, increases the complexity of submitting an RFC to Quilt, by requiring an additional PR to be made after acceptance of most RFCs. Additionally, it is not clear when an RFC will be required for a revision, which might discourage people from contributing revisions to Quilt specifications.
@@ -42,11 +45,8 @@ Each specification only contains the actionable parts of the RFCs involved (usua
 
 # Rationale and Alternatives
 
-- Why is this the best possible design?
-- What other designs are possible and why should we choose this one instead?
-- What other designs have benefits over this one? Why should we choose an
-  alternative instead?
-- What is the impact of not doing this?
+- Having a specifications repository keeps all official documents about policy in one spot.
+- We could continue with the existing process, but having policy mixed between RFCs is not desirable as described in the Motivation section above.
 
 
 # Expected Response
--- a/rfc/0016-specification-repository.md
+++ b/rfc/0016-specification-repository.md
@@ -1,0 +1,54 @@
+# Summary
+
+This RFC amends the RFC process with the creation of a specifications repository, which holds the body of some policy defined through RFCs here.
+
+This RFC does not seek to change the way policy and documentation is reviewed and accepted, only how it is amended and referenced.
+
+# Motivation
+
+Currently, policy on team structure, voting, and other critical aspects of governance are defined only through RFCs.
+This is an issue when a major part of that RFC needs to be redefined--the entire RFC template is excluded for these changes! This is fine for clarifications, but in large-scale changes that require lots of discussion, the current process works against the point of defining these documents through RFCs in the first place.
+
+
+# Explanation
+
+The `quiltmc/specifications` repository created by this RFC can hold three major types of specifications:
+
+1. File specifications
+2. Governance policy
+3. Technical definitions (see: ASMR, QSL)
+   For simplicity, each of these kinds of documents will be called specifications within this RFC.
+   Each specification would be its own file.
+   Creation or major revisions (to be determined by a case-by-case basis) of a specification requires an RFC.
+   Clarifications and other minor revisions can be made by making a PR directly to the specification repository.
+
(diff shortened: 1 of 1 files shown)


rfc/0016-specification-repository.md
```diff
@@ -0,0 +1,54 @@
...
+- Creation of a specification for technical definitions is at the author's discretion, but may be required in cases where major parts of the document are expected to be revised, or if an amendment PR to the RFC repository is deemed as "too big" and needing its own RFC.
+
+# Drawbacks
+
+- This change, if accepted, increases the complexity of submitting an RFC to Quilt, by requiring an additional PR to be made after acceptance of most RFCs. Additionally, it is not clear when an RFC will be required for a revision, which might discourage people from contributing revisions to Quilt specifications.
+- This change assumes that it is not desired for RFCs to be the central authority on policy, but instead to provide a look at the initial motivation and intentions behind a change *before* it is implemented.
+
+# Unresolved Questions
+
+- Should the restrictions on technical definitions be stricter? relaxed? or are they fine?
```
i509VCB, line 43: I'd say leave the conventions of the specification repo to the specification repo. If there is an argument there we can bring it to an RFC.

i509VCB:
Otherwise I think this is fine.

rfc/0016-specification-repository.md
```diff
@@ -0,0 +1,54 @@
...
+
+1. File specifications
+2. Governance policy
+3. Technical definitions (see: ASMR, QSL)
+   For simplicity, each of these kinds of documents will be called specifications within this RFC.
+   Each specification would be its own file.
+   Creation or major revisions (to be determined by a case-by-case basis) of a specification requires an RFC.
+   Clarifications and other minor revisions can be made by making a PR directly to the specification repository.
+
+Each specification only contains the actionable parts of the RFCs involved (usually summary and explaination). Motivation, Rationale, Unresolved Questions, etc. are left only on the RFC.
```
gdude2002, line 25: This whole section reads... oddly. Documents? Sections? Is point three meant to be that huge paragraph? You define "specifications" within point three, but use the terminology before that in the section.

rfc/0016-specification-repository.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Requirements for Creation
+
+- Creation of a specification is required for file specifications and governance policy.
+
+- Creation of a specification for technical definitions is at the author's discretion, but may be required in cases where major parts of the document are expected to be revised, or if an amendment PR to the RFC repository is deemed as "too big" and needing its own RFC.
```
gdude2002, line 31: Is this implying that anyone can create a specification for anything, for any reason? Or is it implying that new specifications don't need to be reviewed? I'm not entirely sure.

gdude2002 (changes requested):
You mentioned on Discord that this relates to the community team - but the community team doesn't use the RFC process for policy work, and this document doesn't explicitly mention the community team, what it does or how this specifically affects it anywhere.

Otherwise, I think it reads OK for now.

rfc/0016-specification-repository.md
TheGlitch76: Point three is not meant to be that long, fixing that.

rfc/0016-specification-repository.md
```diff
@@ -0,0 +1,54 @@
...
+
+# Rationale and Alternatives
+
+- Having a specifications repository keeps all official documents about policy in one spot.
+- We could continue with the existing process, but having policy mixed between RFCs is not desirable as described in the Motivation section above.
+
+
+# Expected Response
+
+This change should make it easier for the community to locate a central authority on policy and specifications, and encourage long-form discussion of major revisions to community policy through the proper RFC process.
```
misode, line 54: > a central authority on policy and specifications

This is what the RFC repository is for. Adding a specification repository would make it confusing because you might not be sure where something is specified. Quilt projects, wiki pages, website pages, etc could all just reference an RFC when they need to.

rfc/0016-specification-repository.md
```diff
@@ -0,0 +1,54 @@
+# Summary
+
+This RFC amends the RFC process with the creation of a specifications repository, which holds the body of some policy defined through RFCs here.
+
+This RFC does not seek to change the way policy and documentation is reviewed and accepted, only how it is amended and referenced.
+
+# Motivation
+
+Currently, policy on team structure, voting, and other critical aspects of governance are defined only through RFCs.
+This is an issue when a major part of that RFC needs to be redefined--the entire RFC template is excluded for these changes! This is fine for clarifications, but in large-scale changes that require lots of discussion, the current process works against the point of defining these documents through RFCs in the first place.
```
misode, line 10: This shouldn't be an issue. If there needs to be made a small tweak, the RFC can be amended. Otherwise a new RFC can be made replacing the old RFC.

misode (commented):
I don't see the usefulness in a separate specification repository. In my opinion it makes it more confusing.

rfc/0016-specification-repository.md
TheGlitch76: How would you handle situations where an RFC is replaced by a new one?

TheGlitch76:
I have updated the RFC to reflect some formatting errors I missed when writing, as well as rewording some sections.

Additionally, letting anyone create a specification for any thing has been removed, and it is now (more or less) clearly defined what should and shouldn't become a specification

rfc/0016-specification-repository.md
misode: The new RFC would mention something like "This standard supersedes the one specified in RFC 123" and a small edit is made to the old RFC like "This RFC is obsoleted by RFC 234". As prior art, here's the first example I came across: https://datatracker.ietf.org/doc/html/rfc2822.

TheGlitch76:
Closing this, I have a better idea for an alternative that uses the RFC repo directly instead.

PR #17, opened 2021-04-25
RFC 0017: Standard Feature Toggle.

LambdAurora:
This RFC aims to defined a standardized way to add feature toggles to mods and synchronize between servers and clients.

[Rendered markdown](https://github.com/LambdAurora/rfcs/blob/standard-feature-toggle/rfc/0017-standard-feature-toggle.md)

### DIFF
--- a/rfc/0017-standard-feature-toggle.md
+++ b/rfc/0017-standard-feature-toggle.md
@@ -1,0 +1,140 @@
+## Summary
+
+This RFC proposes a possible standardized way to synchronize feature toggles (boolean values) between the client and server.
+
+
+## Motivation
+
+Currently there is no easy way for developers to implement feature toggles unless they are using the custom payload packets directly with the networking API which requires more work.
+
+The drawbacks of developers using non-standardized protocols is the impossibility to write a common server mod to disable some features in some mods. This aims to reduce work to multiple different parties.
+
+
+## Explanation
+
+### How would this work?
+
+The idea behind this RFC is for mod developers there would be an API for which mods can register a feature, for a given feature you can set a client value (from the client config), and query if the feature is enabled or not (by checking if the server allows it then it returns the client value, else it returns false).  
+Each features are identified by an ID.  
+Servers can send a packet changing the values of those features by ID.  
+When a client joins a server it would send a or multiple packet(s) describing which features exist on the client.
+
+Mods can set whether the server must explicitely authorize the features to enable it, or explicitely disallow the features to disable it.
+
+### Networking specification
+
+To standardize this feature toggle system it needs a standardized networking specification so anyone can re-implement it with a custom API for different modding platforms (Spigot, Sponge, Forge, etc).
+
+The networking specification is divided in multiple parts:
+ - networking types
+ - networking packets
+
+All types used are either defined in this RFC or are used in MC's own networking. `S->C` means Server to Client and `C->S` means Client to Server.
+
+In this specification, the type `string` without parenthesis describes an UTF-8 string with a maximum length of `32767`.
+
+### Array `A[]`
+
+An array holds a list of values of the specified type.
+It also contains the size `N` of the array.
+
+| Fields | Type     | Description                                 |
+|:------:|:--------:|:--------------------------------------------|
+| size   | [VarI32] | The size of the array.                      |
+| 0      | A        | The first element of the array if present.  |
+| 1      | A        | The second element of the array if present. |
+| ...    | A        | An element of the array if present.         |
+| N-1    | A        | The N-1 element of the array if present.    |
+
+#### FeatureList
+
+| Fields    | Type     | Description                                                     |
+|:---------:|:--------:|:---------------------------------------------------------------:|
+| namespace | string   | The namespace of the features. Can be considered the modid too. |
+| features  | string[] | The features inside the given namespace.                        |
+
+#### FeatureAuthorization
+
+Can be described as an enum with two possible values: `ALLOW` or `DENY`.
+
+ - `ALLOW` means that the client should follow the user configuration's and is allowed to enable the related feature.
+ - `DENY` means that the client is forbidden to enabled the related feature.
+
+In networking this can be represented as a boolean with `ALLOW` being `true` and `DENY` being `false`.
+
+#### Feature
+
+Represents a feature with an authorization value.
+
+Can be seen as a tuple of a string for the identifier path, and a [FeatureAuthorization].
+
+#### FeatureOverrideList
+
+Represents a list of features with authorization values.
+
+| Fields    | Type     | Description                                                     |
+|:---------:|:--------:|:---------------------------------------------------------------:|
+| namespace | string   | The namespace of the features. Can be considered the modid too. |
+| features  | [Feature]\[\] | The features inside the given namespace.                        |
+
+#### `quilt:available_features` packet
+
+Describes the available features on the client to the server. This packet can be split into multiple packets.
+
+Direction: `C->S`
+
+| Fields      | Type              | Description                                                     |
+|:-----------:|:-----------------:|:---------------------------------------------------------------:|
+| expect_more | bool              | Determines whether the server should expect follow-up packets or not. If a `quilt:available_features` packet is sent after a similar packet which had `expect_more` field set to `false`, then it should override the feature list. |
+| features    | [FeatureList]\[\] | The features available on the client.                           |
+
+#### `quilt:features_override` packet
+
+Describes the overriden features on the client by the server. This packet can be split into multiple packets.
+
+Direction: `S->C`
+
+| Fields      | Type                      | Description             |
+|:-----------:|:-------------------------:|:-----------------------:|
+| features    | [FeatureOverrideList]\[\] | The features overrides. |
+
+### Who implements it?
+
+In the case of Quilt, this would likely be its own QSL module, or integrated in the QLS networking module.
+
+In the case of other modding platforms, anyone is free to re-implement the protocol and offer an API.
+
+## Drawbacks
+
+This could add an extra burden to some mod developers and the QSL.
+
+It could also place too much trust into the client which should be avoided as the number one1 rule of client/server relationships is do not trust the client.
+
+
+## Rationale and Alternatives
+
+- This design allows for a generic server-mod to disallow/allow features due to the standardization.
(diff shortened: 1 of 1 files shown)


Scotsguy (changes requested):

rfc/0017-standard-feature-toggle.md
```diff
@@ -0,0 +1,140 @@
...
+#### FeatureList
+
+| Fields    | Type     | Description                                                     |
+|:---------:|:--------:|:---------------------------------------------------------------:|
+| namespace | string   | The namespace of the features. Can be considered the modid too. |
+| features  | string[] | The features inside the given namespace.                        |
+
+#### FeatureAuthorization
+
+Can be described as an enum with two possible values: `ALLOW` or `DENY`.
```
i509VCB, line 58: One question people may ask it how would more granular feature sets be allowed or denied given a more boolean-like state shown here.

i509VCB:
I am going to think about this for a bit - I have some ideas here.

Haven-King:
I intend on having a config API with proper config value syncing. In a world where that exists, it doesn't make much sense to me to implement this as well, when there's so much functional overlap.

LambdAurora:
> I intend on having a config API with proper config value syncing. In a world where that exists, it doesn't make much sense to me to implement this as well, when there's so much functional overlap.

@Haven-King I Believe those two things try to achieve different goals.

A config API would be very good at **_synchronizing_** configurations between the client and the server (both directions I guess), but it's not really designed to allow features or ranges (if this RFC is modified to allow more complex data-structure than just an enum), while this RFC aims directly at allowing features and *maybe ranges* (I currently have no idea how this could be represented on the network).

The other issue is a config API implies that the mod is also present on the server, while this aims to be as generic as possible so you could have a common server mod managing permissions stuff and multiple client mods using this.

Haven-King:
How is a feature or range different from a config value? As for the mod being required on the server, it implies no such thing. Config values are synced with an ID and value much like your features would be, and grants all the flexibility that goes along with that.

zeroeightysix:
@LambdAurora 
> A config API would be very good at synchronizing configurations between the client and the server (both directions I guess), **but it's not really designed to allow features or ranges**

Can you clarify? I'm sure a capable configuration API would have all the necessary features in place to introduce things like ranges and 'feature toggle' metadata.

> The other issue is a config API implies that the mod is also present on the server

I agree with @Haven-King on that this is not at all the case. A reasonable compromise would be to require the **config API** to be present on the server, which I don't believe to be a problem.

I do want to plug [fiber](https://github.com/FabLabsMC/fiber) here, a configuration system Pyrofab and I primarily worked on. It was meant to cover most usecases the fabric ecosystem needed, and grew quite capable of those. Unfortunately it hasn't been very active for the past year, but I'd love to blow new life into it if it turns out suitable as a solution to this RFC (or as a standard configuration system for quilt).

leo60228:
This seems like a very niche feature. I feel like this fits more in individual mods than in QSL.

Haven-King:
@zeroeightysix I have been building out [Conrad](https://github.com/Haven-King/Conrad) as a prototype to a standard config library for Quilt. Pyrofab has looked into it and approves of my general approach, so we'll likely be building off of that.

chexo3:
> In other words, feature override would be akin to enforcing servers' rules by code.

Config sync =/= feature override, and you should not create the tools to “enforce rules via code” with feature blocking that requires client cooperation anyway.

Don’t give server admins a false sense of security.

Hibiii:
It's not meant to replace or even enhance server moderation, cheaters will still cheat regardless, I know that much. This will just make it so people who don't want to cheat don't accidentally cheat. In fact, if you allow me to draw parallels, this is like [HackControl for the MC Classic protocol extension](https://wiki.vg/Classic_Protocol_Extension#HackControl) (hack in this context being modded niceties), it controls if you can have nice things or not (like not flying in a parkour level), but not how nice nice things are (flight speed).

gdude2002:
I think it's a good middle ground for a lot of things - yes, it's not anticheat, and realistically Quilt's job isn't to prevent cheating.

However, there are a lot of things it would be excellent for - since the implementation is left up to the mods themselves, there's a lot of possibilities for a feature like this:

* Mods that disable or enable specific parts of themselves (or the entire thing) according to server settings
* Worlds or world regions on servers that toggle mod features dynamically based on player position
* Ways to allow staff members to use more "cheaty" mod features without requiring them to use a separate modpack

That's just a few things off the top of my head. Try not to think of this with a narrow mindset, because the possibilities really are quite extensive.

chexo3:
I don’t think it should be thought of as a feature toggle though. Really it should be part of config syncing. Mods can support changing config values during gameplay if they want, but they don’t necessarily need to.

It shouldn’t be a general type of thing. The server can already give clients cheaty things like flying and ESP, it doesn’t need a mod client side to do that. In modern Java edition a sort of feature toggle like this would end up just being used to give a false sense of security, and has privacy implications depending on how much data is shared with the server.

Also, tools for this already exist: plugin channels

QSL should have tools for mods to set up plugin channels, but I don’t think we should try to create a “standard” around them or have a plugin channel for feature toggles available by default. Let mods handle syncing or not syncing with the server. If you really want a “feature toggle” API, create it outside of Quilt itself and let mods use it, IMO.

LambdAurora:
> Also, tools for this already exist: plugin channels
> 
> QSL should have tools for mods to set up plugin channels, but I don’t think we should try to create a “standard” around them or have a plugin channel for feature toggles available by default. Let mods handle syncing or not syncing with the server. If you really want a “feature toggle” API, create it outside of Quilt itself and let mods use it, IMO.

This RFC is using plugin channels, and the prior art (LambdaControls) also uses them.
The goal here is to give a *way* simpler way to deal with feature toggles stuff, the main reasoning is to abstract it to avoid having 500 standards of mods doing it, making it a pain for server admins to support.

The goal here is to offer a much more generic solution that is simpler to interact with.

And about the false sense of security, I don't see how this does it?
It's **not** replacing anticheats, it's just giving new tools to both modders and server admins.

falkreon:
Regarding features and security, let's step back and think about what the respective sides should be permitted to do and what they can be expected to do.

**The Client**

Can present features that it is loaded with, which it is unable to disable without a disconnect. These may change basic assumptions about networking and sync behavior, and missing one of these can cause the server to unintentionally crash a client. So this presentation of features by the client is for the client's benefit. I would call this `INSTALLED` features.

We also have the opportunity to offer a set of features which can enhance the relationship between the client and server in some way. Think, "WorldEdit GUI". Plugin channels can probe this data manually by sending a S->C message along the feature's channel and awaiting a response. No response means that the feature can't be used. But I'm thinking about teaching people how to do that custom packet handshake versus teaching people to register a feature, and it's pretty clear which one is more approachable. I'd call these features `OFFERED`

**The Server**

The server can report a list of features which are mandatory to have a smooth and feature-complete experience. Let's call this `REQUIRED` features. This set must be a strict subset of the client's `INSTALLED` + `OFFERED` feature set, otherwise this API should alow the server to give players the most helpful disconnect message it can.

I think it might be helpful, upon learning the client's `OFFERED` features, to report an `ACCEPTED` feature list, which completes an optional feature handshake.

"Deny" makes no sense from a server standpoint. It's probably a net security loss for the server to advertise blacklisted features, as the slightly-more-than-casual cheater will change one letter of the feature name and slide right in, or the feature will simply not advertise itself. If you want to say, "I'm incompatible with this" that's another thing, but that's not the sense I'm getting from this RFC and discussion.


**TL,DR;**

 It makes sense to me for the client to positively enumerate `INSTALLED` and `OFFERED` features, and it makes sense for the server to positively enumerate `REQUIRED` and `ACCEPTED` features, and the names of these are important. It does not make much sense to me for either side to negatively assert anything.

gdude2002:
This sounds fine at a basic level, but feature negotiation should not be limited to the connect phase - it should be possible for a server to say "no more feature X" or "now we're allowing feature X" at any point. This is because it provides a convenient way for mods to support runtime toggling of features without having to make use of custom logic - for example, GOML claims where you can't use your minimap.

anonymous123-code:
From my understanding, `INSTALLED` and `REQUIRED` are compability checks similar to problems regsync tries to solve. To me it seems like those are unchangeable while joined on a server and, similar to the server declaring incompabilities, out of scope for this RFC.

Now, with the `OFFERED` and the `ACCEPTED` features left, this is quite similar in behavior to the current RFC: The client sends a list of features it can offer, and the server responds with the features out of that list it accepts. The one difference, except from the different implementation, is the fact that features automatically get their own plugin channels. However, assuming that we want multiple mods providing the same features, this might encourage mod incompabilities as these will share channels - rendering providing plugin channels useless.

Additionally, I don't see why a deny functionality should be a security loss - malicious mods can simply not make use of this API.

The one security concern I have is that this functionality *might* provide information on the server's anticheat - which cannot really be avoided.

anonymous123-code:
The one concern I have with both concepts is a privacy concern: The client still sends data to the server, probably revealing the mods using this system.

To avoid this, the server simply could send overrides for all features where the defaults differentiate from the server's standards. This should come with some sort of wildcard to catch e.g. all radar features, but then re-enabling one specific sub-feature.

I also wrote about such a system in the archived Thread [#ModID list alternatives in #qsl-general](https://discord.com/channels/833872081585700874/967500834676035685) on the [Quilt Toolchain Discord](https://discord.gg/afZYcQpbwv)
grafik
grafik

falkreon:
The Discord poster has struck an important note here: A lot of the concepts in this RFC aren't entirely separate from the concern of permissions. That's not a good thing!

There are some places where permissions and networking concerns dovetail: when a server communicates permissions to a client (again, as permitted extensions) you can prevent desyncs. You cannot require a client to disable a clientside feature; you can only ask it nicely. But this is *still* a permission concern, not a capability concern.

We absolutely do need a way to negotiate capabilities, but don't get them confused.

SylvKT:
> This appears to be meant for use in quilt only. However, I propose we allow other mod loaders to use the same packet standard. It's not really a "standard" if it's only going to be used in one thing. Now, I see why you would make this; you can technically implement this in alternative APIs, but my point is that if we want compatibility between fabric/forge servers and quilt clients (or vice versa), we'll want to do something like POSIX.

Is this still planned?

PR #24, opened 2021-06-08
Fix issues found in the QMJ specification during implementation

TheGlitch76:
The summary of changes, and their reasoning:
1. Maven groups now have to be alphanumeric or the characters `.`, `-`, or `_` (I am willing to expand this if needed)
2. The maven group namespace `loader.plugin.` is reserved for use by plugins, for example fabric mods would be under the implicit group `loader.plugin.fabric.modid`
     - This isn't enforced for `provides` though, should it be? Or should `provides` not be allowed to define a maven group at all?
3. Provides now uses a `ProvidesObject` instead of a `DependencyObject`--some fields of DependencyObject needlessly complicate implementation or are simply not needed (for example, the array behavior of a dependency object)

### DIFF
--- a/loader/0002-quilt.mod.json.md
+++ b/loader/0002-quilt.mod.json.md
@@ -61,11 +61,11 @@ A unique identifier for the organization behind or developers of the mod. The gr
 A unique identifier for the mod or library defined by this file, matching the `^[a-z][a-z0-9-_]{1,63}$` regular expression. Best practice is that mod ID's are in snake_case.
 
 ### The `provides` field
-| Type             | Required |
-|------------------|----------|
-| DependencyObject | False    |
+| Type  | Required |
+|-------|----------|
+| Array | False    |
 
-A [DependencyObject](#dependency-objects) describing other mods/APIs that this package provides.
+An array of [ProvidesObject](#provides-objectss describing other mods/APIs that this package provides.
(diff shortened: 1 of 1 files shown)


LambdAurora (approved):

Haven-King:
1. Sounds good.
2. `provides` should be allowed to provide a Maven group. It is a relatively infrequently used feature, however, and I believe that if somebody wants to provide a "reserved" namespace they might as well be able to.
3. Could you explain in some detail the need for a separate `ProvidesObject` to me? You replaced the single `DependencyObject` with an array of `ProvidesObject`s in your changes, which makes me wonder why you claim the array behavior of `DependencyObject` is not needed. I could also see a situation arising where `unless` would be useful for `provides`, so am wary to remove that as well.

TheGlitch76:
> which makes me wonder why you claim the array behavior of DependencyObject is not needed. 

The array behavior is "any", not "all", which makes no sense for a `provides` object

Haven-King:
While it might not make sense in that a mod wouldn't provide any of the items in provides, it does correlate better to matching dependencies. That being said, I can understand these changes and if a use case arises in the future where `unless` or the array variant is needed, we can always add it.

Haven-King (approved):

PR #29, opened 2021-07-02
Community amendment 1

gdude2002:
[Rendered View](https://github.com/QuiltMC/rfcs/blob/community-amendment-1/rfc/0007-community-team.md)

PR to address a bunch of changes that need to be made to the community team's governance and structure.

A summary of the changes made within this amendment follows:

* Keyholders are explicitly no longer members of the Community Team, but they're still welcome to oversee the community
  team's decisions and processes and provide feedback and insight if they wish. This also means they're no longer
  considered to be voting members.
* It's no longer required that all Community Team members are moderators. While the majority of the Community Team
  should be moderation and management staff, there's also room for event managers, organisational supports, and so on.
* Trainee Community Moderators have been defined, and new Community Moderators must go through a training period before 
  they can graduate and become Community Moderators. Trainee Community Moderators do not vote on internal polls, and
  are not expected to handle difficult situations.
* The process for handling internal votes has been separated out from the rest of the processes, and standardized. This
  allows for a consistent voting process regardless of what's being voted on, while still allowing for extra closing,
  failing and passing conditions as required. Failing conditions also override passing conditions, instead of the other
  way around.
* The voting process now takes abstentions into account, both explicitly (where the voting member states they're 
  abstaining) and implicitly (where the voting member simply doesn't vote).
* Admins are no longer considered to be voting members unless they're part of the Community Team, or consensus for a
  vote cannot be reached. 
* For Community Team elections, the interview now comes before the vote, rather than the other way around.
* A new process to be followed when existing Community Team members wish to switch their role or take on new roles has
  been written.


rfc/0029-community-amendment-1.md
```diff
@@ -0,0 +1,304 @@
...
+processed on Discord. Quilt's Community Team members are elected using a democratic internal process:
+
+1. Members of Quilt community spaces may apply to join the community by contacting the Community Team directly. The
+   preferred contact method for this is by making use of the ModMail system on Discord.
+   
+   * Alternatively, candidates may be put forward by other members of the Community Team, after some internal 
+     discussion. This may be a result of observations made by the Community Team, suggestions from community members,
+     or otherwise - but the Community Team should always discuss before putting forward a candidate, and they should
+     ensure that the candidate is actually interested in the position they're being put forward for.
+1. Candidates must be interviewed by a Community Manager before the election process can continue. All voting members
```
Laarryy, line 255: Is making it a requirement that a CM *begins* the interview necessary? I don't see how it would be a problem since that's what's been happening, but I'm not sure strictly enforcing it is necessary either.

rfc/0029-community-amendment-1.md
```diff
@@ -0,0 +1,304 @@
...
+* Maintaining informational channels, and writing and posting community-oriented announcements
+* Managing any bots or other community-relevant automation tools
+* Organising, training and supporting the moderation team, and acting as supplemental moderators as required
+* Supporting the rest of the Community Team in their work, providing them with the tools and changes to community 
+  spaces that they need
+
+Community managers must be appointed democratically, following the process outlined later in this document. If
+required, the keyholders will handle handing out the permissions required for community managers to work in Quilt
+community spaces.
+
```
Laarryy, line 118: Perhaps adding a section for what tools/abilities Community Manager, Moderator, Trainee Moderator, and Admin are given is something that should be done? I'll give an example suggestion under Moderators, since that's the only one I'm completely sure I know what tools I have access to (things like banning, modmail, etc)

rfc/0029-community-amendment-1.md
```diff
@@ -0,0 +1,304 @@
...
+  policies
+* Collaborating with other Community Team members to ensure the smooth running of the community, and to ensure that due
+  process is followed - in accordance with the community policies
+* Engaging in active, honest participation in the Quilt community
+* Engaging in the Community Team's democratic processes
+* Helping members to find the correct community space for their query, and ensuring that other members don't prevent
+  them from getting the support they need
+* Keeping up to date on social issues (especially those that affect minority groups), and applying that knowledge to
+  keep community members safe
+
```
Laarryy, line 136:
```suggestion

The tools available to Community Moderators will include both full moderation (ban, kick from community spaces) and ModMail abilities. This is in order to allow Moderators to effectively handle any situation that needs attention, as well as to participate in the appeals and application processes.

```

rfc/0029-community-amendment-1.md
```diff
@@ -0,0 +1,304 @@
...
+All members of the Community Team have the following responsibilities:
+
+* To represent the Quilt project in an honest and respectful manner, and serve as an example of good behaviour in
+  community spaces
+* To participate in the processes that shape the Community Team, including voting on new team members and structural
+  changes to community spaces (unless they're a Trainee Community Moderator - see below)
+* To engage with the community authentically, being present and visible in the community when able and appropriate
+* To be supportive of other Community Team members, providing insight, advice and support when able and appropriate
+* To help with the planning and running of community events and projects, when able and appropriate
+
```
Laarryy, line 79: Suggestion to include here what tools/powers are available to the 'basic' Team. I propose only what is strictly essential for the position. For example, moderation powers will not be required, but in order to participate in electing new team members, ModMail viewing permission at *least* will be necessary.

rfc/0029-community-amendment-1.md
```diff
@@ -0,0 +1,304 @@
...
+All team roles are additive - they add additional responsibilities on top of this basic set.
+
```
Laarryy, line 81: Suggestion for a process by which Team Members that have no other additional roles can apply for them. I can easily imagine scenarios in which a person on the team may be highly suitable for participation in community event planning, but should not be allowed to self-assign the power to ban individuals. This could be because of longstanding rivalries, inability to think fairly and calmly in conflict etc. 

```suggestion

Once an individual is made a part of the Community Team, additional roles can, when the need presents itself, be internally applied for. In order to facilitate this, a new ModMail ticket which is restricted from view of the applicant will be made and the interview process will proceed as if for a newly-entering team member. For example, a member of the Community Team applying to a Community Moderator position would be expected to provide any and all information and possess all abilities that a non-team member would applying for Moderator.

```

This is just a suggestion for a basic procedure, and I would be happy to see my basic idea improved on.

Laarryy (changes requested):
!! Thanks for the work in writing this up!

rfc/0029-community-amendment-1.md
gdude2002: I don't think it's necessarily a hard requirement, but it's something that should be a Manager responsibility rather than a Moderator one - if anyone's gonna accidentally put their foot in their mouth, it should probably be a manager haha

rfc/0029-community-amendment-1.md
gdude2002: I think this is something that's likely to change over time, as we open up new spaces and existing spaces change and get updated - maybe we should list it _somewhere_, but I want to avoid having to create a new RFC amendment every time

rfc/0029-community-amendment-1.md
gdude2002: A promotion track, then? That's not a bad idea, honestly. I'll look at adding one.

gdude2002: I've added one in the latest push.

Laarryy (approved):
Everything looks wonderful!

rfc/0029-community-amendment-1.md
Laarryy: Fair 'nuff

rfc/0029-community-amendment-1.md
```diff
@@ -0,0 +1,325 @@
...
+* The poll's closing date has been reached
+* The poll has failed or passed early due to at least 75% of the voting members of the Community Team voting in that
+  direction, excluding abstaining members from that count
+* The special closing conditions specified as part of the specific process that this vote is part of have been met, if
+  any
+
+### Interpreting the Vote
+
+When a poll has closed, the votes should be inspected. If the final vote has a margin of three votes or fewer, 
+non-voting members of the administrative board must be invited to vote on the poll, with their votes counting for
```
mrmangohands, line 214: > non-voting members of the administrative board must be invited to vote on the poll, with their votes counting for three

If the number of non-voting Admins is even they could also tie, which doesn't seem ideal. Not sure if we want to tweak the process here somehow or leave as is.

mrmangohands (approved):
Generally this is looking good to me aside from the question about Admin votes.

rfc/0029-community-amendment-1.md
gdude2002: The way it currently works is that moving a modmail thread out of the "Backlog" category and into one of the others will prompt you to sync perms. If you decline this, only mods and managers can see it - if you accept, admins can also see it.

Of course, once this RFC is in, we'll need to reorg the community team role structure a bit - but it's the same kind of idea.

rfc/0029-community-amendment-1.md
gdude2002: I'll change this to include admins that have already voted as well, that was an oversight on my part.

gdude2002:
I'm removing the WIP status from this PR, and I'll post it in the meta channel for those that may want to comment.

rfc/0007-community-team.md
```diff
@@ -1,3 +1,12 @@
+# Amendments
```
TheGlitch76, line 1: I'm not a fan of how you have done this kind of amending. This might work fine for the first amendment, but what about two or three years down the line?

I recommend either entirely superseding the RFC (with changes marked), or editing RFC 7 directly and letting version control sort out the history for everyone.

Haven-King: Agreed. That was a large part of the motivation for doing these in a git repo in the first place.

gdude2002: We had no amendment stuff (that I could see) in the RFC process RFC, so I had no idea what was expected really - it'd be simple enough to do that, if that's the right path.

Technici4n: Agreed, the git diff should be enough for history. Viewing the current version in its entirety is the most important thing after all.

SpaceClouds42: I agree, i think the amendments format makes it a bit harder to follow what the current state is. git can take care of showing changes

gdude2002: This has been done.

Redblueflame (approved):

PR #30, opened 2021-07-05
Add a load type enum field

AlexIIL:
[Rich Diff](https://github.com/QuiltMC/rfcs/pull/30/files?short_path=4f0177c#diff-4f0177c28d2ee6074572764a79131c032c2058b924cf0e25872c2547fed9f736)

In order to support *not* loading modules if they aren't needed I'd like to add a new enum field `load_type` to quilt.mod.json. I should have opened this a while ago, sorry. (This is a port of FabricMC/fabric-loader#342)

### DIFF
--- a/loader/0002-quilt.mod.json.md
+++ b/loader/0002-quilt.mod.json.md
@@ -165,11 +165,11 @@ An array of [dependency object](#dependency-objects)s. Defines mods that this mo
 
 Influences whether or not a mod candidate should be loaded or not. May be any of these values:
 
-* "always"
+* "always" (default for mods directly in the mods folder)
 * "if_possible"
 * "if_required" (default for jar-in-jar mods)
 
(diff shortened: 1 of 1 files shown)


AlexIIL:
I marked this as a draft while I tested an implementation for this - which is done now.

i509VCB:
My guess is plugin infrastructure would have to exist or rely on mods correctly declaring their dependencies to handle the loading?

AlexIIL:
We *could* make plugins handle this, but I think it makes more sense to include this in the base mod defining? it's not very complicated functionality anyway.

Haven-King (approved):

TheGlitch76 (approved):

TheGlitch76:
We missed the formal FCP but this PR is a minor change and has sat accepted for two weeks, so we're merging it now.

PR #32, opened 2021-07-08
RFC 32: Quilt Loader Plugins

Haven-King:
[Rendered view](https://github.com/Haven-King/rfcs/blob/loader-plugins/loader/0032-loader-plugins.md)


CheaterCodes (approved):
This is well written and laid out.
I'm a little confused by some of the API structure, but that is out of scope for this RFC anyway.
Looking forward to a less monolithic loader!

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
...
+
+<details>
+    <summary>ITentativeLoadOption</summary>
+    
+```java
+/**
+* {@link LoadOption}s can implement this if they must be processed at the end of the cycle in order to either be
+* added as a normal LoadOption, or removed automatically.
+*/
+public interface ITentativeLoadOption {
```
i509VCB, line 211: Nit: `I` class prefix doesn't align with what I'd say are the agreed but not concrete naming standards at Quilt

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
...
+</details>
+
+<details>
+    <summary>QuiltLoaderPlugin</summary>
+
+```java
+/**
+* @param <T> the types of resolver this plugin can resolve
+*/
+interface QuiltLoaderPlugin<T extends CandidateResolver> extends  {
```
i509VCB, line 148: what does this extend?

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
...
+*
+* If this is an ITentativeLoadOption then it will be removed at the end of the cycle, and handled by whatever plugin 
+* added it.
+*/
+void addOption(LoadOption option);
+
+/**
+* Gets the metadata for a given mod.
+*
+* Because the mods in QuiltLoader only reference fully loaded bods, this method can be used during the modloaded process
```
i509VCB, line 118:
```suggestion
* Because the mods in QuiltLoader only reference fully loaded mods, this method can be used during the mod loading process
```

loader/0032-loader-plugins.md
Haven-King: These were the names Alex gave me, and they work well enough to convey the meaning for now. Very few if any of them are likely to make it into a final implementation 😄

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
...
+idea of how developers might use this feature down the road.
+
+<details>
+    <summary>QuiltPluginContext</summary>
+
+```java
+/**
+* Passed to loader plugins to define what actions they are able to take.
+*/
+public sealed interface QuiltPluginContext permits QuiltPluginContextImpl {
```
i509VCB, line 83: I assume this is just to show this is not for implementation by users. Since loader does still need to support java 8 of course.

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
...
+/**
+* The plugin that this context is for. This method is useless, it just indicates that every other method here has an 
+* implicit paramater of "The Loader Plugin" for the UI / logging to use in some way
+*/
+QuiltLoaderPlugin plugin();
+
+void addCandidate(ModCandidate candidate);
+
+void addCandidate(PluginCandidate candidate);
```
i509VCB, line 92: not sure why we need to have a distinct method for each. What I know at the moment is that a mod may contain a plugin, but a plugin may not contain a mod.

loader/0032-loader-plugins.md
Haven-King: It is, though support for Java 8 is another conversation that should probably be had. I don't think bending over backwards to support versions earlier than 1.17 should be a priority for us.

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
...
+*/
+void addOption(LoadOption option);
+
+/**
+* Gets the metadata for a given mod.
+*
+* Because the mods in QuiltLoader only reference fully loaded mods, this method can be used during the mod loading process
+* to get the metadata for any candidates (tentative or not) that were present prior to this cycle.
+*/
+ModMetadata getMetadata(String modId);
```
i509VCB, line 121: At this point in the loading process we are working with a case where multiple candidates for the mod may exist. Is returning one metadata correct there?

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
...
+    <summary>CandidateResolver</summary>
+    
+```java
+interface CandidateResolver {
+/**
+* Attempts to resolve a tentative candidate.
+*
+* @return an error message if resolution fails, null otherwise
+*/ 
+@Nullable String resolve();
```
i509VCB, line 136: What would resolution do in this context? My interpretation here is things such as downloading mods or remapping a jar, such as with patchwork.

If it's the latter, would it make more sense to return a future?

i509VCB: And with a future, a failed future could be returned given resolution failing.

loader/0032-loader-plugins.md
i509VCB: for the other issue but a note: supporting versions newer than 8 is possible but not the reverse if loader is compiled for java 16

loader/0032-loader-plugins.md
Haven-King: There should only be one candidate for a given mod at this point. This method retrieves metadata for any mods that were added in a *previous* cycle, after sat4j verifies them and discards any superfluous candidates.

loader/0032-loader-plugins.md
Haven-King: A jar may contain both a mod and a plugin, and a quilt.mod.json may define both a mod and a plugin, but they are two distinct entities. For example, a mod candidate can contain bytecode transformations, while a plugin cannot. Plugin classes also must be loaded in time for the next cycle, while mod classes should really not be added to the classpath until we're ready to properly launch. That's why it made sense to me to have two distinctly separate methods here, but we may end up doing something totally different anyways.

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
...
+adding/removing mods and even adding other plugins. Plugins will also have the
+ability to open/create UI elements which can be used for as simple a thing as
+prompting the user for confirmation, or as complex of one as writing a full-scale
+mod launcher as a loader plugin.
+
+## Application
+Loader plugins are initially added from the classpath, allowing default loader
+plugins for an instance to be added via the profile json. All plugins are then
+applied in cycles. During each cycle, the following things happen:
+- Each plugin is given a chance to add any plugin candidates
```
TheGlitch76, line 56: Is this order defined? Does it need to be?

Haven-King: Do you mean the order of plugin application? If so, then no, it is undefined and should not be depended on, similarly to mod load order. Within a cycle, each mod will receive the same inputs. That is, the mod list will not change between plugin applications in the same cycle.

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
...
+*         If this returns false then no rules will be removed, and instead loader will assume that
+*         the error has been handled in some other way. (and it will promptly crash if you haven't)
+*/
+default @Nullable Rule handleOtherErrors(QuiltPluginContext context, List<Rule> errorChain) { return null; }
+
+/**
+* @param dep The dependency which is missing completely. If you can find a valid source for this then you should add 
+*            it with {@link QuiltContext#addTentativeCandidate()}
+* @return True if this plugin did something which will solve / change the error in future,
+*         and so loader won't ask any other plugins to solve this.
```
Chocohead, line 182: With no consistent plugin load order does this not open up the possibility for inconsistent loading behaviour depending on which plugin was fixing errors? Above you mentioned all plugins getting the same information each cycle; would it not be better to find out if there are multiple plugins which all claim to be able to fix an error, in which case you'd either have to pick one consistently or crash? Similar in a way to how cpw's ModLauncher voting doesn't allow all transformers to pass a cycle, at least one must be happy to run every time.

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
...
+and other concerns regarding that will take place at a later date, in another RFC.
+
+
+# Expected Response
+
+Loader plugins have been one of the earliest features Quilt promised to deliver. As
+such, I expect the community will largely be excited to see this feature
+implemented. Projects such as [Bundle Installer](https://github.com/FoundationGames/Bundle-Installer) will be able to make use of loader plugins right from day one, and
+I expect that we'll see a lot of similar projects that aim to make the mod
+installation process as easy as possible for end users.
```
Chocohead, line 328: You already have these now - they're called launchers! In fact we had them even with Modloader, people having trouble managing their mods is nothing new, nor are many of the solutions to trying to fix said troubles. I'd be sceptical of why having multiple similar but different systems is making the process easier than either providing a single uniform, centralised one or just pushing for better launcher integration in the first place.

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
...
+prompting the user for confirmation, or as complex of one as writing a full-scale
+mod launcher as a loader plugin.
+
+## Application
+Loader plugins are initially added from the classpath, allowing default loader
+plugins for an instance to be added via the profile json. All plugins are then
+applied in cycles. During each cycle, the following things happen:
+- Each plugin is given a chance to add any plugin candidates
+    - Each new plugin added this way is given a chance to add any plugin candidates.
+      Rinse and repeat.
```
Chocohead, line 58: Can all plugins see (in classpath terms) all plugins which have loaded up to the point they run? Can they talk between each other directly? Suppose this is linked to the library plugin question

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
...
+
+The addition of Quilt Loader plugins will make Quilt Loader a plugin loader
+first, and mod loader second. Much of the existing mod loading behavior will be
+moved to one or more first-party loader plugins that are installed by default,
+keeping only the bare minimum amount of code required for mod resolving in Loader
+proper. Loader plugins will be capable of effecting the mod loading process by
+adding/removing mods and even adding other plugins. Plugins will also have the
+ability to open/create UI elements which can be used for as simple a thing as
+prompting the user for confirmation, or as complex of one as writing a full-scale
+mod launcher as a loader plugin.
```
Chocohead, line 50: Refraining from the obvious `Curse/Overwolf` -> `Vanilla launcher` -> `Plugin launcher` -> `actually start trying to launch the game` criticism of someone making a mod launcher plugin, is there any encouragement for standardisation of things like UI elements/design? What existing structures/templates (if any, e.g the mod resolution error window) would Loader provide to plugins? Would a plugin which merely acted as a library for other plugins to use be appropriate or would that be better as a pure library that is shaded by plugins which want it?

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
+# Summary
+
+This RFC describes plugins for the Quilt mod loader, and aims to describe the
+purpose(s), functionality, and implications of such a system. At a high level,
+loader plugins will allow developers to control the mod loading process itself,
+giving them access to powerful tools for defining, loading, and removing mods.
```
Chocohead, line 6: I don't see how you'd remove a mod with the API provided, it all looks additive?

Chocohead (changes requested):

TheGlitch76:
I'd like to see Chocohead's comments addressed before this PR is merged.

loader/0032-loader-plugins.md
Haven-King: > would it not be better to find out if there are multiple plugins which all claim to be able to fix an error, in which case you'd either have to **pick one consistently** or crash

This would probably be a better solution, but picking one consistently is the difficult part here. I suppose with loader plugins I wouldn't have much of an issue with adding an integer priority and determining based off that, but a more elegant (and less potentially abusable) system would be preferable.

Haven-King:
> I'd like to see Chocohead's comments addressed before this PR is merged.

I definitely did not forget to click "submit review". That could never happen.

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,328 @@
...
+```
+</details>
+
+
+# Drawbacks
+
+All existing behavior can be implemented using loader plugins. As such, the largest
+drawback may be the increased scope for the Quilt project as a whole, as we will no
+longer be limited to mods specifically. This is less of a drawback and more of an
+organizational change that needs to be considered moving forward.
```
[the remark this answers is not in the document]
TheGlitch76: > plugins should be able to make clear when that happens.

Certainly, but we can't promise that plugins will actually be any good and report their errors properly.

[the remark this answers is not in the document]
Haven-King: We can't _promise_ anything. I don't think asking plugin developers (most of which will likely be Quilt team members) to ensure they have understandable error messages is far-fetched. I don't think this is something we can address directly, but I can definitely add it to the RFC as a potential drawback.

AlexIIL (changes requested):
I don't have any issue with the description, but I do have a few minor concerns for the proposed API surface.

Haven-King:
At this point I am moving this RFC into the final comment period, despite the lack of comments so far. It is likely that much of what has been laid out in this RFC will change throughout development, but this idea is core to a lot of the technical changes Quilt aims to make, so it is crucial that we get work started on this.

The final comment period will close on August 4th.

loader/0032-loader-plugins.md
```diff
@@ -0,0 +1,339 @@
...
+  /**
+   * Gets all currently present load options for mods with the given ID.
+   * If none are present, returns an empty set.
+   */
+  Set<ModLoadOption> getModLoadOptions(String modId);
+
+  /**
+   * Gets all currently present load options.
+   */
+  Map<String, Set<ModLoadOption>> getModLoadOptions();}
```
AlexIIL, line 136: Typo
```suggestion
  Map<String, Set<ModLoadOption>> getModLoadOptions();
```

AlexIIL (approved):

mschae23:
In the "API Surface" part, you're using `Future`s as parameters that will be called.

If that is `Future` from Java, that's not correct. Futures can't be called, they're just delayed suppliers.

I'd suggest a custom functional interface for that, in this case.

skyrising (approved):
I'm also currently using a very hackish way of loading LiteLoader mods along side fabric-loader ([LiteFabric](https://github.com/skyrising/LiteFabric/blob/main/src/main/java/de/skyrising/litefabric/impl/PreLaunch.java)) so I'm very excited for a proper solution to this.

PR #25, opened 2021-06-15
Quilt Loader PR Policy

TheGlitch76:
[Rendered View](https://github.com/theglitch76-forks/rfcs/blob/loader-pr-policy/loader/0000-pr-policy.md)

PR #35, opened 2021-07-19
Quilt Loader PR Policy

TheGlitch76:
I messed up my fork so this needs to be reopened. Sorry.
Comes from #25 

[Rendered View](https://github.com/theglitch76-forks/rfcs/blob/loader-pr-policy/loader/0035-loader-pr-policy)


Leo40Git (approved):
Your rendered view is broken because the RFC file mysteriously lost its `.md` extension.  
Otherwise, this is an excellent policy!

PR #37, opened 2021-08-02
Create 0037-quilt-mappings-structure.md

OroArmor:
Outlines how the Quilt Mappings repository and releases should be done.


rfc/0037-quilt-mappings-structure.md
```diff
@@ -0,0 +1,51 @@
...
+### Branches
+
+* Branches for every major release version of Minecraft, with the initial commit be just the changes from the previous major version.
+* A single branch for development.
+
+The releases branch would publish to the releases maven, while snapshots would be published to the snapshot maven. This would help people scanning the maven find the latest version easier.
+
+### Versioning
+
+With the development branch and release branches, it might be possible to have consistent versioning across Minecraft versions, so the gradle dependency name could be `org.quiltmc:quilt-mappings:${mappings-version}-${minecraft-version}`
```
ByMartrixx, line 34: Are we going to use semver for the mappings version?

rfc/0037-quilt-mappings-structure.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Pull Requests
+
+New mappings will always happen to the `main` branch, and should be reviewed by at least 3 members of the mappings team. (This number can change depending on how many people are on the teams, this is the number Fabric used).
+Pull requests should have the `snapshot` or `refactor` target. 
+* The `snapshot` pull requests should only happen during snapshot phases of Minecraft's development, and will only change mappings introduced between major versions.
+* The `refactor` pull requests can happen at any time, and should only update mappings from the current and previous versions of Minecraft. This will help increase consistency between names across versions. If a refactor wants to change names that follow the new changes, but are in only one of the versions, another pull request should accompany it.
```
ByMartrixx, line 41: > another pull request should accompany it.

Is the other pull request to apply the same changes to the previous version without including the new changes? If that is the case, would it target the previous version branch?

ByMartrixx (approved):

rfc/0037-quilt-mappings-structure.md
mschae23: If Quilt Mappings is going to follow semver, `mappings_version` and `minecraft_version` should be swapped.

OroArmor: yeah i just checked the spec and it would be 
```suggestion
With the development branch and release branches, it might be possible to have consistent versioning across Minecraft versions, so the gradle dependency name could be `org.quiltmc:quilt-mappings:${minecraft-version}+${mappings-version}`
```

rfc/0037-quilt-mappings-structure.md
mschae23, line 41: If some mappings get out of date during snapshots, will you have to create two pull requests now (one for new mappings, and one for changing the outdated ones)?

OroArmor: say `m_asdf` is in 1.16.5 and `m_fdsa` is in 1.17. they are both named `badName`. someone wants to refactor to `goodName`. is this what you are asking?

OroArmor: two pull requests would be needed, but also know that with hashed mojmap any mojang rename or added parameters would loose the mappings, which is somewhat intended. However, I should add a section about pulling mappings with hashed changes or parameter changes to the same method

mschae23: > know that with hashed mojmap any mojang rename or added parameters would loose the mappings

Ah, right. Forgot about that.

An example would be `MixedNoisePoint`, which kept its name in 1.18, but isn't actually a point anymore. So if someone would create a PR with 1.18 mappings, they would have to open a separate PR for this refactor.

But with hashed Mojmap (and small refactor PRs probably getting merged rather quickly), this shouldn't be a problem.

OroArmor: So correction, adding parameters does not lose the name, so only one pull request would be needed in the example above.

I think I understand what you are saying about opening a separate PR, and it would not necessarily be a refactor as the experimental snapshots arent part of the normal game update path.

ByMartrixx:
[Rendered view](https://github.com/OroArmor/rfcs/blob/quilt-mappings-structure/rfc/0037-quilt-mappings-structure.md)

LambdAurora (commented):
This definitely looks like it overcomplicates the entire mappings structure.

Not having a branch for each version (development or release) is confusing, and tags won't help solve that. I fear that this scares potential contributors as it's not as simple as "clone, edit branch, push to fork, PR" anymore.

OroArmor:
Yeah. I wanted to see what a different approach might bring, but the simplicity of yarns current approach is unbeatable.

OroArmor:
Closing because yarns current layout and process is fine.

PR #38, opened 2021-08-11
Tweak provides & arrays in breaks to make more sense.

AlexIIL:
Providing a range of versions means we have to compare version ranges with each other, which makes little sense.

Currently an array in a breaks field is no different to just specifying them separately, unlike for dependencies - so this tweaks that behaviour to allow specifying a break if all of the matching mods are found. For example:

Example A:
````json
"breaks": [
 [ "better-than-buildcraft", "buildcraft-transport" ],
 "buildcraft-silicon"
]
````
Example B:
````json
"breaks": [
 "better-than-buildcraft",
 "buildcraft-transport",
 "buildcraft-silicon"
]
````

In the current schema both example A and B would have the same effect - if any of the three mods are present then the game won't load. In this proposed schema then example A would only fail to load if both "better-than-buildcraft" and "buildcraft-transport" were present.

### CODE REVIEW

**Comment by AlexIIL** [comment]
One alternative to the proposed schema would be to alternate between "ANY" and "ALL" in each layer of array - starting at ANY for depends, and ALL for breaks. I don't think it would make the implementation any more complicated, but it might make displaying it a bit more difficult.

**Comment by Haven-King** [comment]
> One alternative to the proposed schema would be to alternate between "ANY" and "ALL" in each layer of array - starting at ANY for depends, and ALL for breaks. I don't think it would make the implementation any more complicated, but it might make displaying it a bit more difficult.

I think that such behavior would be seldom used and very unexpected to somebody who hasn't read the specification. For that reason alone, I don't think it's necessary, and that the changes proposed in the PR itself are adequate.

**Review by Haven-King — APPROVED** [review]

**Comment by i509VCB** [comment]
Alternation between any and all feels confusing imo.

**Comment by AlexIIL** [comment]
I've updated this to clarify that I'm *not* changing the behaviour of "unless" dep objects, for deps or breaks.

**Review by TheGlitch76 — APPROVED** [review]

**Review by Leo40Git — APPROVED** [review]

**Comment by AlexIIL** [comment]
Sorry, I didn't realise this had conflicts :(

**Comment by AlexIIL** [comment]
Ah, fortunately it was just a file move. This should be mergeable now.

File: specification/0002-quilt.mod.json.md [hunk 1]
```diff
@@ -439,12 +439,12 @@ Defines the identifier and optionally version range of another mod that this pac
...
-| Type         | Required | Default |
-|--------------|----------|---------|
-| Array/String | False    | `"*"`   |
+### The `version` field
+| Type   | Required |
+|--------|----------|
+| String | False    |
 
-Should be a [version specifier](#version-specifier) or array of version specifiers defining what versions this dependency applies to. If an array of versions is provided, the dependency matches if it matches ANY of the listed versions.
+Should be a valid version to provide. if omitted then this defaults to the version of the mod.
```
EnnuiL, line 447:
```suggestion
Should be a valid mod version. If omitted, then this defaults to the version of the providing mod.
```
The original text felt very confusing. This should solve the problem

EnnuiL (changes requested):
The example quilt.mod.json and the description for provide objects should be updated in order to reflect the change from version ranges to versions (respectively: `"versions": "^1.0.0"` to `"version": "1.0.0"`, and `optionally version range` to `optionally version`)

specification/0002-quilt.mod.json.md
AlexIIL: Sure, thanks!

PR #40, opened 2021-09-21
Create proposal

OroArmor:


Leo40Git (approved):

BasiqueEvangelist:
Why not just use Mapping IO?

OroArmor:
1. Mappings IO has a different scope than this proposed project. 
2. Mappings IO is not in house. I would like something in house. As well, the rfc covers how the projects will still be published separately, but be integrated in one larger project to reduce code duplication.

LambdAurora (approved):

TropheusJ:
I don't think not in house is a good excuse to not use something. If it works, we should use it. I can't really give input on that though since I don't know what exactly mappings IO does.

Otherwise, this looks like a good plan to me.

AlsoNiko (approved):

AlsoNiko:
If code is constantly being duplicated between tools I see no reason not to have them as one project instead

TheGlitch76:
> Mappings IO is not in house. I would like something in house. As well, the rfc covers how the projects will still be published separately, but be integrated in one larger project to reduce code duplication.

This isn't a good justification for not using something. Has mappings-io been investigated? We don't have to keep merging with upstream on it, but have we considered its strengths and weaknesses, and if it could possibly be used in the building of tiny utils?

NoComment1105 (approved):
In my opinion reducing the amount of duplicated code, as well as reducing the amount of different pieces of software used would be good

ByMartrixx (commented):
I think Fabric is changing to use Mapping IO everywhere, for example Loom and Lorenz Tiny have migrated, while in Enigma and Yarn there is an issue/pr to do so. We could use Mapping IO as a starting point for Tiny Utils

PR #45, opened 2021-12-16
Revise community team removal process

mrmangohands:
Changes the removal process for community team members not to require a counterargument before acting, and also explicitly outlines that there may be a private place for proposal, discussion, and voting. I think the community team needs the ability to act swiftly and flexibly here, and the staff in question still have an opportunity to respond and request the decision be overturned after the fact. Downside is that they don't necessarily get to know or respond before the result is publicly visible.

### DIFF
--- a/structure/0007-community-team.md
+++ b/structure/0007-community-team.md
@@ -425,7 +425,10 @@ Eligible voters are decided in allignment with
 [RFC 6: Governmental Structure](/structure/0006-governance.md#removal-of-staff) (anyone above or below), but a
 counterargument is not required or expected before voting or removal of staff permissions. Instead the call to action,
 discussion, and voting can occur in a private channel/dm visible to eligible voters, and if it succeeds the removed
-staff can submit a counterargument asking the decision be overturned after the fact.
+staff can submit a counterargument asking the decision be overturned after the fact. If there are concerns about
+damage to the community, project, or team, a Community Manager or the Key Holder may temporarily revoke staff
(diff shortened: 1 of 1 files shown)


Akarys42:
I'd like to see a sentence explicitly allowing a community manager to temporarily remove privileges from the CT member while the vote is ongoing. Other than that, that change seems good :+1:

gdude2002:
Agreed, I think this works but it should be documented that a manager can remove permissions from the CT member in question while the vote is ongoing for cases where leaving the perms in place could cause damage to the community, project, or community team

mrmangohands:
Updated, added the key holder as well for completeness sake. Also I put a deadline of a week for initiating a vote after removing perms, in most cases it should be more like a few hours but erred on the longer side in case there's a situation that requires waiting for more info.

gdude2002 (approved):
I think this should work

Akarys42 (approved):
Lgtm

structure/0007-community-team.md
```diff
@@ -421,8 +421,14 @@ during this process.
 
 ## Process: Removing Team Members
 
-Removal of members from the community team follows the process outlined in 
-[RFC 6: Governmental Structure](/community/0006-governance.md#removal-of-staff).
+Eligible voters are decided in allignment with
```
wafflecoffee, line 424: Typo, "alignment"

wafflecoffee (approved):
LGTM besides the one typo

PR #50, opened 2022-02-21
Change `environment` field to use `dedicated_server` instead of `server`

TheGlitch76:
Just some cleanup as we prepare Loader for the beta.

I've seen people get confused about how the `environment` field refers to the physical dedicated server and not the logical server, and marking their mods that work fine on an integrated server with the `server` environment. This change will hopefully help reduce this error.

@AlexIIL (the only other active Loader team dev at this time) has signed off on this and this is a trivial change that will be implemented immediately, so this needs little-to-no FCP.


LambdAurora (approved):

P03W (approved):

OroArmor (approved):
Good clarification.

Leo40Git (approved):

AlexIIL (approved):

ByMartrixx (approved):

PR #51, opened 2022-03-21
QSL Versioning.

LambdAurora:
This document describes the QSL versioning scheme to address QuiltMC/quilt-standard-libraries#79.

[Rendered view](https://github.com/LambdAurora/rfcs/blob/0051/specification/0051-qsl-versioning.md)

### DIFF
--- a/specification/0051-qsl-versioning.md
+++ b/specification/0051-qsl-versioning.md
@@ -36,11 +36,34 @@ Meaning that the root library version should also have its `major` number increm
 The `minor` number should always start at `0`, and will increment every time a module/library is updated with backwards-compatible changes. It is reset when the `major` number changes.
 
 The `patch` number should always start at `0`, and will increment every time a module/library is updated to fix bugs in a backwards-compatible manner.
-This should not cover the case of a port as [the metadata](#metdata-content) already gives enough information.
+This should not cover the case of a port as [the metadata](#metadata-content) already gives enough information.
 
-### Version qualifier
+> Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence.
+>
+> from [semver](https://semver.org/#spec-item-10)
 
+This means that to ensure proper solving by Quilt loader, <abbr title="Quilt Standard Libraries">QSL</abbr> modules must depend on exactly one Minecraft version.
 
+### Version qualifiers
+
+Now, with the values of the main three version numbers out of the way, we have to carefully consider how we approach version qualifiers.
+
+We will divide this section in different cases:
+
+- __Stable release:__  
+  stable releases should have no qualifiers, thus for the first ever release we would have `1.0.0` as the version string (ignoring [the metadata for now](#metadata-content)).
+- __Beta release:__  
+  a beta release is a release which is not yet stable, its unstability means any breaking change still can happen and the module/library is still under development.
+  Though it also means the module/library is close to full release, it starts to stabilize: the goal is to detect any defects before actual release.
+  Thus the version should be qualified as beta, betas are versioned separately with `-beta.<beta_number>`, `beta_number` starts at `1` and increases for every new beta of the designated release.
+  Beta releases are not available through the stable <abbr title="Quilt Standard Libraries">QSL</abbr> Maven artifact, or through a stable library version, but they are available through the beta version of the artifact.
+  A Beta artifact may depend on stable components if there's no betas for those components currently.
+- __Experimental release:__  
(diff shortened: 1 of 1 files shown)


specification/0051-qsl-versioning.md
```diff
@@ -0,0 +1,70 @@
+# Quilt Standard Libraries Versioning
+
+## Summary
+
+This document intends to specify the versioning scheme of the Quilt Standard Libraries to ensure consistent versioning, reduce confusion and address the case of experimental and beta versions.
+
+## Motivation
+
+As <abbr title="Quilt Standard Libraries">QSL</abbr> gets closer to a Beta release,
+it is important for it to figre out a versioning scheme.
```
Akarys42, line 10:
```suggestion
it is important for it to figure out a versioning scheme.
```

specification/0051-qsl-versioning.md
```diff
@@ -0,0 +1,70 @@
...
+When a module or library is created, the `major` number should always be `1` at the start.
+It mays increment in the case of breaking change, whether it is caused by the <abbr title="Quilt Standard Libraries">QSL</abbr> team or by an update to Minecraft.
+
+If a module is updated with a breaking change, the library that includes it should also increment its `major` number.
+Meaning that the root library version should also have its `major` number incremented.
+
+The `minor` number should always start at `0`, and will increment every time a module/library is updated with backwards-compatible changes. It is reset when the `major` number changes.
+
+The `patch` number should always start at `0`, and will increment every time a module/library is updated to fix bugs in a backwards-compatible manner.
+This should not cover the case of a port as [the metadata][#metdata-content] already gives enough information.
```
Akarys42, line 39:
```suggestion
This should not cover the case of a port as [the metadata](#metdata-content) already gives enough information.
```

Akarys42 (changes requested):
Two typos

specification/0051-qsl-versioning.md
```diff
@@ -0,0 +1,70 @@
...
+When a module or library is created, the `major` number should always be `1` at the start.
+It mays increment in the case of breaking change, whether it is caused by the <abbr title="Quilt Standard Libraries">QSL</abbr> team or by an update to Minecraft.
+
+If a module is updated with a breaking change, the library that includes it should also increment its `major` number.
+Meaning that the root library version should also have its `major` number incremented.
+
+The `minor` number should always start at `0`, and will increment every time a module/library is updated with backwards-compatible changes. It is reset when the `major` number changes.
+
+The `patch` number should always start at `0`, and will increment every time a module/library is updated to fix bugs in a backwards-compatible manner.
+This should not cover the case of a port as [the metadata](#metdata-content) already gives enough information.
```
TheGlitch76, line 39: > the metadata already gives enough information.

From a technical perspective, the metadata actually gives **no information**; [see here](https://semver.org/#spec-item-10):
> Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence.

In other words, if Loader is given `0.1.0+21w11a` and `0.1.0+21w12a`, and both mods are compatible with the current Minecraft version, it is random which will be loaded.

TheGlitch76: > It is defined which one will be loaded using the depends section, if one depends on 21w22a, and the other one depend on 21w11a strictly, and we're loading 21w22a, then the 21w22a version will be picked up.

Then direct in this RFC for QSL to always depend exactly on 1 version.

> Also doesn't semver define lexicographical order for metadata?

I quoted the section where it explicitly excludes metadata from that in my first comment.

LambdAurora: Gotta have my eyes checked I guess.

Unresolved question found! 
Is depending on exactly 1 version desirable?

specification/0051-qsl-versioning.md
```diff
@@ -0,0 +1,107 @@
...
+the metadata takes the targetted version and only keeps the major and minor numbers, disarding the patch number **unless** huge breaking changes happened in that patch.
+In the case of a snapshot, the targetted version is the version the snapshot is for.
+
+## Unresolved Questions
+
+- Is depending on exactly one Minecraft version desirable?
+
+<!-- URLs -->
+[semver]: https://semver.org/ "Semantic Versioning 2.0.0 specification"
+[Fabric API]: https://github.com/FabricMC/fabic "Fabric API repository"
```
DioEgizio, line 103: typo here

```suggestion
[Fabric API]: https://github.com/FabricMC/fabric "Fabric API repository"
```

PR #52, opened 2022-04-02
RFC 2 amendment: add `intermediate_mappings` field to `quilt.mod.json`

TibiNonEst:
An amendment to RFC 2 per https://github.com/QuiltMC/quilt-loader/issues/59.

Adds an optional string field to `quilt.mod.json` in the `quilt_loader` section named `intermediate_mappings`. This string will be used to determine which mappings a mod uses once hashed support is added to QL. The two values it will accept are `"hashed"` and `"intermediary"`.

### DIFF
--- a/specification/0002-quilt.mod.json.md
+++ b/specification/0002-quilt.mod.json.md
@@ -197,7 +197,7 @@ A list of Maven repository URL strings where dependencies can be looked for in a
 |--------|----------|------------------------|
 | String | False    | `"org.quiltmc:hashed"` |
 
-The intermediate mappings used for this mod. The intermediate mappings string must be a valid maven coordinate and match the `^[a-zA-Z0-9-_.]+:[a-zA-Z0-9-_.]+$` regular expresson. This field currently only officially supports `org.quiltmc:hashed` and `net.fabricmc:intermediary`.
+The intermediate mappings used for this mod. The intermediate mappings string must be a valid maven coordinate and match the `^[a-zA-Z0-9-_.]+:[a-zA-Z0-9-_.]+$` regular expression. This field currently only officially supports `org.quiltmc:hashed` and `net.fabricmc:intermediary`.
 
 ### The `metadata` field 
 | Type   | Required |
--- a/specification/0002-quilt.mod.json.md
+++ b/specification/0002-quilt.mod.json.md
@@ -193,11 +193,11 @@ If this mod is in another mods "depends" field then it will be loaded, otherwise
 A list of Maven repository URL strings where dependencies can be looked for in addition to Quilt's central repository.
 
 ### The `intermediate_mappings` field
-| Type   | Required | Default    |
-|--------|----------|------------|
-| String | False    | `"hashed"` |
+| Type   | Required | Default                |
+|--------|----------|------------------------|
+| String | False    | `"org.quiltmc:hashed"` |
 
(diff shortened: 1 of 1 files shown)


specification/0002-quilt.mod.json.md
```diff
@@ -191,6 +192,13 @@ If this mod is in another mods "depends" field then it will be loaded, otherwise
 
 A list of Maven repository URL strings where dependencies can be looked for in addition to Quilt's central repository.
 
+### The `intermediate_mappings` field
+| Type   | Required |
+|--------|----------|
+| String | False    |
```
TheGlitch76, line 198: include the default value here as in other parts of this document

TheGlitch76 (changes requested):

TheGlitch76 (approved):
This will be merged along with the PR to Loader

zml2008:
How about using maven coordinates for this, to have a global identifier for mappings, such as `net.fabricmc:intermediary` and `org.quiltmc:hashed`. This would allow for other namespaces to be contributed, say by loader plugins -- if someone wanted to do say a custom intermediary for legacy versions, or use srg.

TibiNonEst:
My initial idea was that loader plugins could register additional intermediate mapping identifiers such as `"legacy"` or whatever such an identifier may be, although I think using maven coordinates may be better for guaranteeing uniqueness and may make loader plugins easier to create down the line.

chexo3:
> I agree, the maven coordinates would be better for a general approach.
> 
> Should we bother providing aliases matching what was initially proposed here (reserved just for hashed and fabric intermediary), or always use the maven coordinates?

If we were already using this code in production it would make sense, but in this case we can do it without having to worry about old mods, so aliases might be kind of pointless. I think in this case being specific makes sense.

AlexIIL (approved):

TheGlitch76:
Can an admin merge this, please?

specification/0002-quilt.mod.json.md
```diff
@@ -191,6 +192,13 @@ If this mod is in another mods "depends" field then it will be loaded, otherwise
 
 A list of Maven repository URL strings where dependencies can be looked for in addition to Quilt's central repository.
 
+### The `intermediate_mappings` field
+| Type   | Required | Default                |
+|--------|----------|------------------------|
+| String | False    | `"org.quiltmc:hashed"` |
+
+The intermediate mappings used for this mod. The intermediate mappings string must be a valid maven coordinate and match the `^[a-zA-Z0-9-_.]+:[a-zA-Z0-9-_.]+$` regular expresson. This field currently only officially supports `org.quiltmc:hashed` and `net.fabricmc:intermediary`.
```
wafflecoffee, line 200:
```suggestion
The intermediate mappings used for this mod. The intermediate mappings string must be a valid maven coordinate and match the `^[a-zA-Z0-9-_.]+:[a-zA-Z0-9-_.]+$` regular expression. This field currently only officially supports `org.quiltmc:hashed` and `net.fabricmc:intermediary`.
```

OroArmor:
I'll merge when the last comment is resolved.

wafflecoffee:
@OroArmor You can just commit it yourself, no? I didn't mean for that to be a blocking comment, just didn't want a typo

OroArmor:
You're right I can.

PR #47, opened 2021-12-28
RFC 0006: Governance Amendment

CheaterCodes:
Rendered Views:
- [Governence RFC](https://github.com/CheaterCodes/rfcs/blob/governance-amendment/structure/0006-governance.md)
- [Technical Teams RFC](https://github.com/CheaterCodes/rfcs/blob/governance-amendment/structure/0018-technical-teams.md)

The following changes were made;
- Removal of technical teams
- Removal of technical leads
- Addition of team leads
- Generalization of team responsibilities
- Reformating of team creation RFC template
- Definition of sub teams
- Redefinition of the administrative board
- Change of admin election process
- Incomplete redefinition of processes

I'd like to see more done for standard processes, possible more in line with the [Community Team RFC](https://github.com/QuiltMC/rfcs/blob/master/structure/0007-community-team.md).

PR #55, opened 2022-04-23
RFC 55: Mappings team restructure

ByMartrixx:
This RFC aims to define the new mappings team structure

[Rendered view](https://github.com/ByMartrixx/rfcs/blob/rfc55/structure/0055-mappings-team.md)

### DIFF
--- a/structure/0055-mappings-team.md
+++ b/structure/0055-mappings-team.md
@@ -1,0 +1,70 @@
+# Mappings team restructure
+
+## Summary
+
+The mappings team was originally created in [RFC 18][RFC18], but has since
+evolved: a triage team was created for [Quilt Mappings][QM], and the need to
+split the mappings team into smaller subteams has appeared. This RFC describes
+the proposed structure for the top-level team and those subteams.
+
+
+## Motivation
+
+The scope of the repositories maintained by the main mappings team is very big,
+from the main Quilt Mappings repository, to all of the toolchain used and 
+related to it. By splitting the mappings team, its members can be assigned to
+projects more specifically, and review requests in PRs would only reach to the
+repository maintainers rather than the whole team.
+
+
+## Explanation
+
+The mappings team would be split into three sub-teams: Quilt Mappings, Mappings
+Tooling, and the existing Mappings Triage sub-team.
+
+The Mappings Triage team is only in charge of reviewing mappings submitted to
+QM, ensuring their quality and making sure they stick to the conventions.
+This team only has triage access to the QM repository.
+
+The Quilt Mappings team would be in charge of maintaining QM, and also shares
+the responsabilities of the triage team. The main difference between the triage
+team and this one is their access level to the repository, meaning only QM
+members can merge PRs or push commits directly.
+
+The Mappings Tooling team would be in charge of maintaining all of the mappings
+toolchain (ie. [Mappings Hasher][Hasher], [Enigma], [Tiny Remapper][TR], etc.),
+and subsequently of reviewing PRs to these repositories.
+
+
+## Drawbacks
+
+Having different people maintain the toolchain and QM could mean more
+bureaucracy to make changes in both QM and the toolchain at the same time.
+
+
+## Rationale and Alternatives
+
+- Having multiple sub-teams instead of one big team can make assigning people
+  to tasks easier, who manages each project clearer
+- Even splitting the mappings team into QM and tooling, the number of projects
+  handled by the tooling team is very broad. The tooling team could be split
+  into more sub-teams
+  - Most of the projects are in charge of one or two people, the sub-teams
+    wouldn't be bigger than 2 people in most cases.
+- Currently, all of the responsabilities mentioned are already handled by the
+  mappings team or the triage team.
+
+
+## Unresolved Questions
+
+- Who should be part of the top-level team?
+- What permissions should top-level team members have in QM, and the toolchain
+  repositories?
+- Should triage team members be considered as Quilt Developers?
+
+
+[RFC 18]: 0018-technical-teams.md
+[QM]: https://github.com/QuiltMC/quilt-mappings
+[Hasher]: https://github.com/QuiltMC/mappings-hasher
+[Enigma]: https://github.com/QuiltMC/enigma
+[TR]: https://github.com/QuiltMC/tiny-remapper
--- a/structure/0055-mappings-team.md
+++ b/structure/0055-mappings-team.md
@@ -27,13 +27,13 @@ QM, ensuring their quality and making sure they stick to the conventions.
 This team only has triage access to the QM repository.
 
 The Quilt Mappings team would be in charge of maintaining QM, and also shares
-the responsabilities of the triage team. The main difference between the triage
-team and this one is their access level to the repository, meaning only QM
+the responsibilities of the triage team. The main difference between the triage
+team and the QM team is their access level to the repository, meaning only QM
 members can merge PRs or push commits directly.
 
 The Mappings Tooling team would be in charge of maintaining all of the mappings
 toolchain (ie. [Mappings Hasher][Hasher], [Enigma], [Tiny Remapper][TR], etc.),
-and subsequently of reviewing PRs to these repositories.
+and subsequently for reviewing PRs to these repositories.
 
 
 ## Drawbacks


OroArmor (approved):

structure/0055-mappings-team.md
```diff
@@ -0,0 +1,70 @@
...
+
+The mappings team would be split into three sub-teams: Quilt Mappings, Mappings
+Tooling, and the existing Mappings Triage sub-team.
+
+The Mappings Triage team is only in charge of reviewing mappings submitted to
+QM, ensuring their quality and making sure they stick to the conventions.
+This team only has triage access to the QM repository.
+
+The Quilt Mappings team would be in charge of maintaining QM, and also shares
+the responsabilities of the triage team. The main difference between the triage
```
wafflecoffee, line 30:
```suggestion
the responsibilities of the triage team. The main difference between the triage
```

structure/0055-mappings-team.md
```diff
@@ -0,0 +1,70 @@
...
+team and this one is their access level to the repository, meaning only QM
```
wafflecoffee, line 31:
```suggestion
team and the QM team is their access level to the repository, meaning only QM

```

structure/0055-mappings-team.md
```diff
@@ -0,0 +1,70 @@
...
+members can merge PRs or push commits directly.
+
+The Mappings Tooling team would be in charge of maintaining all of the mappings
+toolchain (ie. [Mappings Hasher][Hasher], [Enigma], [Tiny Remapper][TR], etc.),
+and subsequently of reviewing PRs to these repositories.
```
wafflecoffee, line 36:
```suggestion
and subsequently for reviewing PRs to these repositories.

```

wafflecoffee (commented):
>  - Who should be part of the top-level team?

I don't really think there should be anyone on the top level, except maybe a Technical Lead per #47. Rather that this should just be split up into the three different responsibilities.

> - Should triage team members be considered as Quilt Developers?

Perhaps I'm biased as a triage team member, but I would say yes. Cheater made a good point in #mappings-internal in saying that QM Triage is very similar to QSL sub-teams, so really there shouldn't be a second class of developers for only QM triage.

ix0rai (approved):

structure/0055-mappings-team.md
```diff
@@ -0,0 +1,70 @@
+# Mappings team restructure
+
+## Summary
+
+The mappings team was originally created in [RFC 18][RFC18], but has since
+evolved: a triage team was created for [Quilt Mappings][QM], and the need to
+split the mappings team into smaller subteams has appeared. This RFC describes
```
gdude2002, line 7: This wording needs some work.

```
but our needs have changed since then - a 
[Quilt Mappings][QM] triage team was created, 
and it became necessary to split the mappings team 
into smaller subteams.
```

structure/0055-mappings-team.md
```diff
@@ -0,0 +1,70 @@
...
+the proposed structure for the top-level team and those subteams.
+
+
+## Motivation
+
+The scope of the repositories maintained by the main mappings team is very big,
+from the main Quilt Mappings repository, to all of the toolchain used and 
+related to it. By splitting the mappings team, its members can be assigned to
+projects more specifically, and review requests in PRs would only reach to the
+repository maintainers rather than the whole team.
```
gdude2002, line 17: More wording work needed here.

```
The Mappings Team maintains numerous projects - from the 
main Quilt Mappings repository, to all the related tooling. This
is quite a large development surface, and splitting the team into
smaller subteams would ensure that tasks will only be assigned
to those that they're relevant to, preventing unnecessary noise
generated by notifications going to people that don't need them.
```

structure/0055-mappings-team.md
```diff
@@ -0,0 +1,70 @@
...
+This team only has triage access to the QM repository.
+
+The Quilt Mappings team would be in charge of maintaining QM, and also shares
+the responsibilities of the triage team. The main difference between the triage
+team and the QM team is their access level to the repository, meaning only QM
+members can merge PRs or push commits directly.
+
+The Mappings Tooling team would be in charge of maintaining all of the mappings
+toolchain (ie. [Mappings Hasher][Hasher], [Enigma], [Tiny Remapper][TR], etc.),
+and subsequently for reviewing PRs to these repositories.
```
gdude2002, line 36: This section appears to have mixed tenses, which makes reading harder than it should be. Additionally, the terminology is inconsistent - is it named `Mappings Team` or `Quilt Mappings Team`?

I'd also avoid abbreviating the team name to `QM Team` to avoid confusion with the QM project, which has the same abbreviation.

Additionally, all team names should be capitalized - `Mappings Team`, not `mappings team`

structure/0055-mappings-team.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+## Drawbacks
+
+Having different people maintain the toolchain and QM could mean more
+bureaucracy to make changes in both QM and the toolchain at the same time.
```
gdude2002, line 42: You don't explain how this is a drawback.

structure/0055-mappings-team.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+## Rationale and Alternatives
+
+- Having multiple sub-teams instead of one big team can make assigning people
+  to tasks easier, who manages each project clearer
```
gdude2002, line 48: I'd clarify this a little by saying `and would help clarify who is in charge of each project`. Also, I'd avoid passive language like `can`, and replace it with more assertive language like `would`.

structure/0055-mappings-team.md
```diff
@@ -0,0 +1,70 @@
...
+- Even splitting the mappings team into QM and tooling, the number of projects
+  handled by the tooling team is very broad. The tooling team could be split
+  into more sub-teams
```
gdude2002, line 51: The wording here is confusing. How about something like this?

```
While splitting the Mappings Team into the Quilt Mappings and 
Mappings Tooling teams would help to split the workload, there 
would still be a large number of projects assigned to the Mappings 
Tooling Team - but this split would allow us to further split the Mappings
Tooling team into further sub-teams
```

However, this begs the question - why is this subdivision not part of this RFC as well, if you're already thinking about it?

structure/0055-mappings-team.md
```diff
@@ -0,0 +1,70 @@
...
+  - Most of the projects are in charge of one or two people, the sub-teams
+    wouldn't be bigger than 2 people in most cases.
```
gdude2002, line 53: This wording doesn't make sense. Perhaps you wanted something more like this?

```
In this situation, most projects belonging to sub-teams would be run 
by one or two people, meaning each sub-team would be quite small
```

structure/0055-mappings-team.md
```diff
@@ -0,0 +1,70 @@
...
+- Currently, all of the responsabilities mentioned are already handled by the
+  mappings team or the triage team.
```
gdude2002, line 55: This doesn't seem to make sense as a point in this section.

structure/0055-mappings-team.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+## Unresolved Questions
+
+- Who should be part of the top-level team?
+- What permissions should top-level team members have in QM, and the toolchain
+  repositories?
+- Should triage team members be considered as Quilt Developers?
```
gdude2002, line 63: I'd remove `as` here. Also, this is a discussion that needs to happen before this RFC is merged, in my opinion.

structure/0055-mappings-team.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+[RFC 18]: 0018-technical-teams.md
```
gdude2002, line 66: You probably wanted to add a `References` header here.

gdude2002 (changes requested):
I've suggested a number of wording changes here. I'd also like to point out a few things in general:

* Team names should be capitalized (`Mappings Team`), and you need to use precisely the same terminology for each team throughout the document
* Don't abbreviate team names (`QM Team`) when those abbreviations aren't defined earlier (formally or informally), or when they conflict with other abbreviations (`QM` to mean the Quilt Mappings project)
* This is a formal document, and it's important to maintain writing consistency:
	*  There's a lot of mixing of tenses: pretty much everything in this document should be present or future tense, but not a mixture
	* The language used is often not assertive enough: "can", "may", "could" vs "will", "would"
	* When writing lists, the punctuation at the end of list items should be consistent throughout the list - have it or don't have it, but don't mix
* This is a Markdown document, and you do not need to wrap lines. Your margin is set especially short, and this makes reviewing and adding suggestions to review comments rather annoying. Instead, please use an editor that soft-wraps at the margin.

I think that's about it for now. Thanks for writing this!

PR #56, opened 2022-04-23
RFC 0056: Permit multiple version constraints in a single version specifier.

AlexIIL:
The quilt.mod.json specification currently only permits a single version constraint pattern per version specifier, which means you can only use a single version comparison per dependency object. While a dependency object's `versions` field permits an array instead of a single string, it is compared using "ANY" logic - so you could depend on (for example) exactly minecraft 1.18.1, or exactly 1.18.2 (`"versions": [ "=1.18.1", "=1.18.2" ]`).

Where this stops being useful is when you want a larger version range, for example between minecraft `1.14` and `1.17` - a naive way of writing this would be `"versions": [ ">=1.14", "<1.17" ]`. Unfortunately this doesn't work, since that actually depends on any version of minecraft. The current way around this is to use *two* nearly identical dependency objects - for example:

```json
"depends": [
  {
    "id": "minecraft",
    "versions": ">=1.14"
  },
  {
    "id": "minecraft",
    "versions": "<1.17"
  }
]
```
However the above format is quite long, and would be even worse if additional fields are used in the dependency object, since they would have to be added to both.

This RFC proposes a more clear syntax for "ANY" vs "ALL" behavior for version specifiers:

```json
"depends": [
  {
    "id": "minecraft",
    "versions": { "all": [">=1.14", "<1.17"] }
  }
]
```

An additional change is that quilt-loader will validate version ranges to ensure they make sense - so accidentally depending on "any version" without using `*`, or accidentally *disallowing* every version will be caught early. (Since we have the `breaks` object to handle this).

## Risks

~~Failing to load quilt mods with invalid version specifiers may break existing quilt mods that have invalid or redundant specifiers, which is unfortunate. However I'd rather we broke them *now* (and informed them that their version specifiers don't work like how they expect) than accidentally allow invalid/redundant versions now and in the future.~~

A previous version of this PR included logic to break mods which used the old format. Since mods have been around for a long time by now, and I don't want to force everyone to use dependency overrides just to make their mods work, this PR no longer breaks mods which might not be specified correctly.


AlexIIL:
As a side note, this will require us to (at least partially) break quilt-loader's API to implement. However I think it's a useful change, and few mods probably use quilt's dependency apis yet (if any), so it shouldn't hurt *too* much.

LambdAurora (approved):

AlexIIL:
After a brief discussion on discord I'm going to re-do this using more explicit language, rather than having arrays have a different meaning to separated strings.

lukebemish:
Any updates on what this RFC needs or where it could be taken up? A way to specify version constraint intersections is currently a bit of a limitation in the `quilt.mod.json` format; is there a discussion that was had about where this should go or a PR that supersedes this?

AlexIIL:
I've *finally* updated this to use `any` and `all` directly, rather than arrays vs spaces in separators. As such this is ready for review again!

specification/0002-quilt.mod.json.md
```diff
@@ -398,9 +398,32 @@ A mod identifier in the form of either `mavenGroup:modId` or `modId`.
...
+
+Should be a single [version specifier](#version-specifier) defining the versions this dependency applies to.
+
+#### Array (Deprecated)
+
+Should be an array of [version specifiers](#version-specifier) defining the versions this dependency applies to. The dependency matches if it matches ANY of the listed versions.
+
+This usage is discouraged as many mod developers wrongly assumed this was "ALL" rather than "ANY", leading to unfortunate situations where mods depend on [ ">=1.18.1", "<1.19.3" ], which actually matches every version of minecraft, rather than just the versions between 1.18.1 and 1.19.2.
+Quilt-loader will emit a formatting warning when discovering mods that use this.
+Quilt-loader will not emit errors if this is used.
```
Leo40Git, line 419:
```suggestion
Quilt Loader will emit a formatting warning when discovering mods that use this.
Quilt Loader will not emit errors if this is used.
```

specification/0002-quilt.mod.json.md
```diff
@@ -398,9 +398,32 @@ A mod identifier in the form of either `mavenGroup:modId` or `modId`.
 ### The `versions` field
 | Type         | Required | Default |
 |--------------|----------|---------|
```
Leo40Git, line 400:
```suggestion
| Type                | Required | Default |
|---------------------|----------|---------|
```

specification/0002-quilt.mod.json.md
```diff
@@ -398,9 +398,32 @@ A mod identifier in the form of either `mavenGroup:modId` or `modId`.
...
+
+#### String
+
+Should be a single [version specifier](#version-specifier) defining the versions this dependency applies to.
+
+#### Array (Deprecated)
+
+Should be an array of [version specifiers](#version-specifier) defining the versions this dependency applies to. The dependency matches if it matches ANY of the listed versions.
+
+This usage is discouraged as many mod developers wrongly assumed this was "ALL" rather than "ANY", leading to unfortunate situations where mods depend on [ ">=1.18.1", "<1.19.3" ], which actually matches every version of minecraft, rather than just the versions between 1.18.1 and 1.19.2.
```
Leo40Git, line 417:
```suggestion
This usage is discouraged as many mod developers wrongly assumed this was "ALL" rather than "ANY", leading to unfortunate situations where mods depend on [ ">=1.18.1", "<1.19.3" ], which actually matches _every_ version of Minecraft, rather than just the versions between 1.18.1 and 1.19.2.
```

Leo40Git (approved):
LGTM, apart from some slight nitpicks.

gdude2002:
Before I review this - how does this PR tie in to #64?

specification/0002-quilt.mod.json.md
```diff
@@ -396,11 +396,34 @@ A dependency object defines what mods/plugins a given mod depends on or breaks.
...
+| Type                | Required | Default |
+|---------------------|----------|---------|
+| Array/Object/String | False    | `"*"`   |
 
-Should be a [version specifier](#version-specifier) or array of version specifiers defining what versions this dependency applies to. If an array of versions is provided, the dependency matches if it matches ANY of the listed versions.
+A version specifier, or complex set of version specifiers that control what versions match this dependency object.
+
+It is an error to specify multiple constraints that conflict with each other (where no version would match the whole specifier).
+
+It is an error if the resulting version specifier matches every version, and isn't the single string `"*"`, or uses the deprecated array style of definition.
```
gdude2002, line 407: You probably want `when` here, not `if`.

specification/0002-quilt.mod.json.md
```diff
@@ -396,11 +396,34 @@ A dependency object defines what mods/plugins a given mod depends on or breaks.
...
+Should be a single [version specifier](#version-specifier) defining the versions this dependency applies to.
+
+#### Array (Deprecated)
+
+Should be an array of [version specifiers](#version-specifier) defining the versions this dependency applies to. The dependency matches if it matches ANY of the listed versions.
+
+This usage is discouraged as many mod developers wrongly assumed this was "ALL" rather than "ANY", leading to unfortunate situations where mods depend on [ ">=1.18.1", "<1.19.3" ], which actually matches _every_ version of Minecraft, rather than just the versions between 1.18.1 and 1.19.2.
+Quilt Loader will emit a formatting warning when discovering mods that use this.
+Quilt Loader will not emit errors if this is used.
+Future schema versions will likely remove this array functionality.
```
gdude2002, line 420: These last three lines should probably be an unordered list, no?

gdude2002 (changes requested):
Just a couple small changes.

specification/0002-quilt.mod.json.md
AlexIIL: Yes, thanks!

specification/0002-quilt.mod.json.md
AlexIIL: Yeah, they are in that kindof order.

AlexIIL:
I consider this "good enough" to be moved to `Final Comment Period`, and this will obsolete #64.

PR #57, opened 2022-06-04
Configuration metadata API

kvverti:
[Rendered](https://github.com/kvverti/quilt-rfcs/blob/master/specification/0000-common-configuration-api.md)

### DIFF
--- a/specification/0000-common-configuration-api.md
+++ b/specification/0000-common-configuration-api.md
@@ -199,15 +199,16 @@ benefits.
 
 ## Rationale and Alternatives
 
-- Why is this the best possible design?
-- What other designs are possible and why should we choose this one instead?
-- What other designs have benefits over this one? Why should we choose an
-  alternative instead?
-- What is the impact of not doing this?
-
 This RFC describes a common API between mods that define configuration and mods that consume configuration.
 This approach gives the most control to mods that wish to consume multiple mods' configuration, e.g. to construct a config screen,
 while avoiding including too many "batteries" that not all mods may use.
+
+### Alternatives
+
+Define an API for constructing config UI elements that is completely separate from the underlying config. This would give mods more freedom to choose
+specific UI representations.
+
+Do nothing. In the absence of a config API, mods will likely tend toward using fully featured GUI libraries for displaying configuration.
 
 ## Prior Art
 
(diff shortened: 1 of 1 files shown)


Leo40Git (approved):
Looks good so far! Unfortunately, I'm not well-versed enough in the technical side of things to critique possible implementations.

specification/0057-configuration-metadata-api.md
```diff
@@ -0,0 +1,230 @@
...
+The most appropriate Widget Type for a given Setting depends on the Setting's underlying data type and on any constraints a Setting's
+value must obey.
+
+If a Setting's Widget Type is `UNORDERED_COLLECTION` or `ORDERED_COLLECTION`, additional data is stored `WIDGET_ELEMENT_TYPE`.
+
+```java
+MetadataType<WidgetType, ?> WIDGET_ELEMENT_TYPE;
+```
+
+#### Value Converter
```
Haven-King, line 80: I also don't think this is necessary. `WIDGET_TYPE` should just be a `MetadataType<Function<TrackedValue<?>, Widget>>`, with different kinds of widgets having either constructors or helper methods for different parameter types. `ToggleWidget` for instance could have the following method:
```java
public static Widget create(TrackedValue<?> value) {
  Object defaultValue = value.value();

  if (defaultValue instanceof Boolean) {
    return new ToggleWidget(value.value());
  } else if (defaultValue instanceof Enum e && e.values().length == 2) {
    return new EnumToggleWidget(value.value());
  } else if (...) {
  
  } else {
    throw new WidgetCreationException(...); // Might be better to have some sort of result class instead of throwing errors, but ¯\_(ツ)_/¯ 
  }
}
```
The value would then simply call
```java
builder.metadata(WIDGET_TYPE, builder -> builder.factory(ToggleWidget::create))
```

specification/0057-configuration-metadata-api.md
```diff
@@ -0,0 +1,230 @@
...
+
+A Group Validator is used to validate properties that pretain to the relationship among multiple Settings. For example, two Settings that
+represent the minimum and maximum value of a range may have a Group Validator that ensures the maximum is always greater than or equal to
+the minimum. It is intended that Group Validators that perform the same validation compare equal so that consumers of this API do not
+run validation logic multiple times.
+
+#### Environment Policy
+
+```java
+public enum EnvironmentPolicy {
```
Haven-King, line 176: `CLIENT_ONLY` and `SERVER_ONLY` are functionally identical. I would therefore rename this class to `SyncPolicy` or something along those lines and condense those two options to `NONE`, with `SERVER_SYNCED` becoming `S2C` and `CLIENT_SYNCED` becoming `C2S`. Additionally, I might add a `P2P` sync type that shares user configs with other clients in a way that allows them to be accessed on a per-player basis. This has a lot of valuable implications for user-set display information (nicknames come to mind).

Haven-King (changes requested):
This spec certainly covers a lot of the bases, but I feel that it overcomplicates some things while leaving others out.

PR #60, opened 2022-06-18
RFC 60: Quilt Mod Menu

Siuolthepic:
[Rendered View](https://github.com/Siuolthepic/rfcs/blob/master/rfc/0060-mod-menu.md) This RFC proposes an official Mod Menu for Quilt alongside an accompanying team.

### DIFF
--- a/rfc/0060-mod-menu.md
+++ b/rfc/0060-mod-menu.md
@@ -1,0 +1,76 @@
+
+## Summary
+
+This document outlines the process for creating an official Mod Menu for usage with Quilt alongside an accompanying team that would maintain it. This mod menu will act as a way to view loaded mods, change configs, read credits, and get links to contacts from within Minecraft.
+
+
+## Motivation
+   As of right now, we do not supply an official Mod Menu for Quilt. The current solution is TerraformerMC's Mod Menu for Fabric, and that does not fully support our additions (Full Quilt support, Quilt Config, quilt.mod.json spec improvements, etc.). Having a standardized Mod Menu would allow for support for these features. Furthermore, it would also be an officially endorsed mod, which would help us lead players to it, and we at The Quilt Project would be able to ensure its quality.
+
+
+## Explanation
+
+### Integration
+A toggle in the Quilt Installer will allow for this mod to be installed at the same time as loader
+
+### Why are we doing this?
+Native Quilt Support, which lets us add additional features from things like quilt.mod.json or Quilt Config without issue.
+Standardization, allowing an official solution compared to Fabric's unofficial Mod Menu
+
+### Why not a fork of Terraformer's Mod Menu?
+The author of the Mod Menu himself, Prospector, said that ModMenu is "janky" - it is likely not worth the effort to clean up code AND integrate with other Quilt Standards. 
+
+### How will we deal with Fabric Mods that require configs?
+We will supply the Mod Menu API to allow for backwards compatibility with configs in Fabric mods.
+
+
+## Drawbacks
+
+### Potentially Unnecessary
+   It maybe be considered to be unnecessary to make this as the current Mod Menu solution on Fabric works fine. It also may complicate the user experience to have multiple mod menu mods. 
+   
+Fabric's Mod Menu is a separate mod, however, so it can simply be removed if a user finds it distracting.
+
+### Removes User Choice
+
+Some people may not want to be forced upon a Mod Menu in that case we are taking away their choice on whether or not to have a Mod Menu. 
+    
+This, however, is almost a non-issue; mod menus are often lightweight, and they are often the best way to abstract access to a mod's config. 
+
+### Burden of Maintenance
+
+Some people may view this as yet another Libarary / Mod to maintain every update and push new features to. 
+
+This is probably the most serious issue - an official Quilt Mod Menu will require official support - that means PR triage, updating between versions, and ensuring that as changes occur to things like the QMJ, the Mod Menu is appropriately updated. We propose the creation of a new team, the Mod Menu team, that would be responsible for its management.
+
+
+
+## Rationale and Alternatives
+
+
+- This allows for features from Loader and QSL to be used within the mod more effectively, and it also allows for easier communication within our teams in designing necessary features
+- As for alternatives, we could wait for a Quilt Mod Menu to show up, we could (against the author's recommendation) fork the TerraformersMC Mod Menu (and still have to create a new team for support). The proposed path is the best option.
+
+
+## Prior Art
+
+- TerraformerMCs Mod Menu
+    - Benefits
+        - Very well designed UI. Its visually pleasing and easy to use.
+        - API to allow for support of other config systems
+        - Supports showing parent/JiJed mods
+    - Drawbacks
+        - No native Quilt support. Thus it cant have support for features in Quilt such as support for Loader Plugin loaded mods, Quilt Config, Contributor Roles.
+        - Its downloaded and installed manually, causing confusion amongst new users
+
+- Forge's Mods screen
+    - Benefits
+        - No installation needed, thus making it more standard and causing less confusion amongst users
+    - Drawbacks
+        - Not particularly well-designed - can be a bit confusing to deal with.
+        - Does not support alternative config systems.
+        - Does not support showing parent/JIJed mods
+
+
+## Unresolved Questions
+- Specific details about the creation of the Mod Menu Team 


rfc/0060-mod-menu.md
```diff
@@ -0,0 +1,76 @@
...
+
+
+## Motivation
+   As of right now, we do not supply an official Mod Menu for Quilt. The current solution is TerraformerMC's Mod Menu for Fabric, and that does not fully support our additions (Full Quilt support, Quilt Config, quilt.mod.json spec improvements, etc.). Having a standardized Mod Menu would allow for support for these features. Furthermore, it would also be an officially endorsed mod, which would help us lead players to it, and we at The Quilt Project would be able to ensure its quality.
+
+
+## Explanation
+
+### Integration
+A toggle in the Quilt Installer will allow for this mod to be installed at the same time as loader
```
OroArmor, line 14: How will updates be sent out? This feature of mod downloads at installation needs to be expanded in my opinion.

CheaterCodes: Updating is way out of scope for this and shouldn't be necessary.

CheaterCodes:
Regarding the Integration section:
This topic is much more complicated than a single sentence could possibly explain. Frankly, this needs a separate RFC outlining a team responsible for maintaining "official quilt mods", of which mod menu would be a subteam.
It would be that teams responsibility to figure out how to integrate quilt mods into the installation process.

RedstoneParadox:
Would there be any official mods other than mod menu, though?

rfc/0060-mod-menu.md
Siuolthepic: After discussing with Finn and Cheater, we decided the integration should probably be its own RFC. I'll be removing it from this one

Siuolthepic:
QSL has it's own team and sub teams. It's not going to be merged into one giant superteam when it's not that necessary

What other official mods could there even be? Official Crash Resumer? Official Iris/Sodium? Official Block Mod? Official Dimension Mod? Mod menu you can make the case that it is a highly essential mod given that it's the main way to configure mods, but what else besides that would be sensible to maintain under a banner of the Quilt Official Mods?

I just really think if your gonna do installation through installer it shouldn't require making a whole superteam just for that. We already have installer deligated anyways

Siuolthepic:
Also, if I am misunderstanding your proposal feel free to say so, it just doesn't make sense to me the way you worded it

Lorenzo0310200 (approved):

rfc/0060-mod-menu.md
```diff
@@ -0,0 +1,73 @@
+
+## Summary
+
+This document outlines the process for creating an official Mod Menu for usage with Quilt alongside an accompanying team that would maintain it. This mod menu will act as a way to view loaded mods, change configs, read credits, and get links to contacts from within Minecraft.
+
+
+## Motivation
+   As of right now, we do not supply an official Mod Menu for Quilt. The current solution is TerraformerMC's Mod Menu for Fabric, and that does not fully support our additions (Full Quilt support, Quilt Config, quilt.mod.json spec improvements, etc.). Having a standardized Mod Menu would allow for support for these features. Furthermore, it would also be an officially endorsed mod, which would help us lead players to it, and we at The Quilt Project would be able to ensure its quality.
```
Prospector, line 8: > Full Quilt support, Quilt Config, quilt.mod.json spec improvements

It supports all of the features except config AFAIK?

rfc/0060-mod-menu.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+## Explanation
+
+### Why are we doing this?
+Native Quilt Support, which lets us add additional features from things like quilt.mod.json or Quilt Config without issue.
```
Prospector, line 14: > Native Quilt Support, which lets us add additional features from things like quilt.mod.json

Native Quilt support is not required to support Quilt features

rfc/0060-mod-menu.md
```diff
@@ -0,0 +1,73 @@
...
+Standardization, allowing an official solution compared to Fabric's unofficial Mod Menu
+
+### Why not a fork of Terraformer's Mod Menu?
+The author of the Mod Menu himself, Prospector, said that ModMenu is "janky" - it is likely not worth the effort to clean up code AND integrate with other Quilt Standards.
```
Prospector, line 18: I'll elaborate here. Mod Menu was created as one of the first Fabric mods and as time has gone on, it has increased in scope and capability. Admittedly, a lot of the code was rather haphazardly thrown together, and many remnants of the original code is visible, especially in the rendering part. The mod was also refactored to support multiple loaders more easily at some point, which also doesn't really make sense for a native mod. There are tons of undocumented features, such as localizable mod descriptions, which you would not know about unless you happened to find it in the code (I myself forgot I added the feature until I came across it in the code once).

I have advised that an official Quilt mod list screen be written from the ground up, however I will say that I am fairly happy with the UI that I have created as I put great effort into making the mod fit as seamlessly into vanilla as possible. I welcome anyone to take inspiration from or straight up rip-off my UI.

rfc/0060-mod-menu.md
```diff
@@ -0,0 +1,73 @@
...
+### Potentially Unnecessary
+   It maybe be considered to be unnecessary to make this as the current Mod Menu solution on Fabric works fine. It also may complicate the user experience to have multiple mod menu mods. 
+   
+Fabric's Mod Menu is a separate mod, however, so it can simply be removed if a user finds it distracting.
+
+### Removes User Choice
+
+Some people may not want to be forced upon a Mod Menu in that case we are taking away their choice on whether or not to have a Mod Menu. 
+    
+This, however, is almost a non-issue; mod menus are often lightweight, and they are often the best way to abstract access to a mod's config.
```
Prospector, line 35: I am not really clear on how the implementation of this would work, but I would suggest that it still be optional? For example, if I create a highly customized and integrated modpack, I would not want there to be a mod menu (and users should not need to access configs anyway).

rfc/0060-mod-menu.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Prior Art
+
+- TerraformerMCs Mod Menu
+    - Benefits
+        - Very well designed UI. Its visually pleasing and easy to use.
+        - API to allow for support of other config systems
+        - Supports showing parent/JiJed mods
+    - Drawbacks
+        - No native Quilt support. Thus it cant have support for features in Quilt such as support for Loader Plugin loaded mods, Quilt Config, Contributor Roles.
```
Prospector, line 60: > Thus it cant have support for features in Quilt such as support for Loader Plugin loaded mods, Quilt Config, Contributor Roles.

This isn't true, contributor roles are already supported and it's not particularly difficult to support more Quilt features without being a quilt-native mod.

rfc/0060-mod-menu.md
```diff
@@ -0,0 +1,73 @@
...
+        - Its downloaded and installed manually, causing confusion amongst new users
```
Prospector, line 61: Arguably can also be a benefit

rfc/0060-mod-menu.md
```diff
@@ -0,0 +1,73 @@
...
+
+- This allows for features from Loader and QSL to be used within the mod more effectively, and it also allows for easier communication within our teams in designing necessary features
+- As for alternatives, we could wait for a Quilt Mod Menu to show up, we could (against the author's recommendation) fork the TerraformersMC Mod Menu (and still have to create a new team for support). The proposed path is the best option.
+
+
+## Prior Art
+
+- TerraformerMCs Mod Menu
+    - Benefits
+        - Very well designed UI. Its visually pleasing and easy to use.
```
Prospector, line 56: Seems a little subjective, maybe instead mention what makes it well designed, such as being consistent with the vanilla game's user interface to give users an intuitive interface.

rfc/0060-mod-menu.md
```diff
@@ -0,0 +1,73 @@
...
+        - No native Quilt support. Thus it cant have support for features in Quilt such as support for Loader Plugin loaded mods, Quilt Config, Contributor Roles.
+        - Its downloaded and installed manually, causing confusion amongst new users
+
+- Forge's Mods screen
+    - Benefits
+        - No installation needed, thus making it more standard and causing less confusion amongst users
+    - Drawbacks
+        - Not particularly well-designed - can be a bit confusing to deal with.
+        - Does not support alternative config systems.
+        - Does not support showing parent/JIJed mods
```
Prospector, line 69: Does forge have these?

rfc/0060-mod-menu.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+## Unresolved Questions
```
Prospector, line 72: How will it be distributed? Is it a team under the QSL project or a separate project entirely? Will it be optional?

rfc/0060-mod-menu.md
```diff
@@ -0,0 +1,73 @@
...
+    - Drawbacks
+        - No native Quilt support. Thus it cant have support for features in Quilt such as support for Loader Plugin loaded mods, Quilt Config, Contributor Roles.
+        - Its downloaded and installed manually, causing confusion amongst new users
+
+- Forge's Mods screen
+    - Benefits
+        - No installation needed, thus making it more standard and causing less confusion amongst users
+    - Drawbacks
+        - Not particularly well-designed - can be a bit confusing to deal with.
+        - Does not support alternative config systems.
```
Prospector, line 68: I don't think this is true at all? I'm pretty sure you can supply your own config GUI just like Mod Menu using a ConfigGuiFactory https://github.com/MinecraftForge/MinecraftForge/blob/0e66b7b21757e510aef706ca320881b9eb268822/src/main/java/net/minecraftforge/client/ConfigGuiHandler.java#L19

rfc/0060-mod-menu.md
duplexsystem: It does as of very recently

rfc/0060-mod-menu.md
Juuxel: Yep, example usage here: https://github.com/Juuxel/Adorn/blob/fdb3e57c65668f27af08411a17e8fccbc7538de4/forge/src/main/kotlin/juuxel/adorn/platform/forge/client/AdornClient.kt#L22-L24

(Adorn uses the same custom config screen with both Mod Menu and Forge)

rfc/0060-mod-menu.md
zOnlyKroks: Atm we are thinking of having it as an option for the installer where u can toggle the installation and such. Updates will be provided via curseforge, modrinth etc.

zOnlyKroks: There will maybe be a team which manages all these content mods, but this is out of scope for this rfc.

rfc/0060-mod-menu.md
zOnlyKroks: We can prob integreate a config option for removing the button from the main menu.

maximumpower55 (approved):
LGTM

OroArmor (approved):
Overall seems somewhat shakily put together and not 100% thought through. While I agree this is something that Quilt should explore, I don't think this RFC fully explores the different options that Quilt can take. Ill approve since I believe that Quilt should put work into this, but I feel like this RFC will likely be mended in the near future.

peaceheis:
I don't have any particular changes to recommend, but I agree with Pro. I think think RFC needs some revising, but I believe the underlying problem is that it's really just invoking a lot of questions that haven't been answered - who is responsible for Quilt content mods? Who is responsible for distribution? How do we distribute them? As time goes on and on, it will become more obvious that this "by-the-seat-of-our-pants" method of adding new things to The Quilt Project will not work - its an organizational hazard. I don't think thats necessarily in the scope of this RFC, but it's definitely something to think about. While it's true The Quilt Project is in beta, I do think we need to consider this organization sooner, rather than later.

peaceheis (approved):

zOnlyKroks:
I will have to discuss the first 2 questions. But I can answer the third. The way we distribute them is via the installer and modrinth and all the other sites. We have an option in the quilt installer where u can choose the mods u like to install. Updates will be provided over curseforge and the other sites there are.

Loweredgames:
And still open? I was also thinking of adding an internal marketplace (free and in moderation) if this proposal is approved. 

Please let me know :)

zOnlyKroks:
Is it open: Yes I suppose so
Is it stale: Yes

PR #64, opened 2022-08-27
RFC 0064: Remove cheatsheet link

AlexIIL:
@OroArmor added this about a week ago, however it references more functionality than what we actually support - namely hyphenated ranges and combined ranges (and/or). In addition I'd like to support this in a slightly different way (see #56), so this cheatsheet link is inaccurate.

### DIFF
--- a/specification/0002-quilt.mod.json.md
+++ b/specification/0002-quilt.mod.json.md
@@ -467,8 +467,6 @@ A version range specifier can make use of any of the following patterns:
 * `~1.0.0` — Matches the most recent version greater than or equal to version 1.0.0 and less than 1.1.0
 * `^1.0.0` — Matches the most recent version greater than or equal to version 1.0.0 and less than 2.0.0
 
-See [this cheatsheet](https://devhints.io/semver) for more information. 
-
 # Drawbacks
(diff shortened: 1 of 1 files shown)


AlexIIL:
To be honest that spec isn't a *bad* thing to support, but it's adding to the original QMJ spec - and so it should be inlined (in effect) into the spec, rather than just linked to.

OroArmor (approved):

gdude2002 (approved):

gdude2002:
The change in this PR is also present in #56, so I think this one should be closed in favour of the other PR.

SilverAndro (approved):

AlexIIL:
This should be merged immediately, as it removes functionality that was never supported or discussed properly.

#56 is a discussion of potential functionality, so it should go through the normal amount of discussion before being merged.

gdude2002:
OK, but the same change is in both PRs - if that's the case then you should remove the change from the other PR so it will merge correctly, no?

AlexIIL:
Closing in favor of #56.

Akarys42:
Superseeded by #56

PR #65, opened 2022-10-02
RFC 62: Quilt Community Collab

Akarys42:
This RFC defines the Quilt Community Collab space.

Any help to spruce up the RFC would be much appreciated.

[Rendered view](https://github.com/Akarys42/rfcs/blob/62-quilt-community-collab/structure/0062-quilt-community-collab.md)

### DIFF
--- a/structure/0062-quilt-community-collab.md
+++ b/structure/0062-quilt-community-collab.md
@@ -46,7 +46,7 @@ Managers are responsible for keeping the Program up-and-running. They take care 
 
 * Maintaining, organizing, and actioning changes to the Program's Discord server, and any other Program-related spaces
 * Organizing polls and votes, as required by formal procedures
-* Aiding in conflict resolution between communities and Program members
+* Aiding in conflict resolution between communities and Program members, and acting as moderators of Program spaces when necessary
 * Staying in touch (and maintaining relations) with the communities that are part of the Program
 * Procuring and maintaining any services that the Program needs to function properly
 
@@ -70,7 +70,9 @@ The Event Planner role is given to members of participating communities' communi
 
 ### Project Roles
 
-Informal roles may be created and granted to Program members that are contributing to (or participating in) Program-related projects. The definition of these roles will depend on the project in question, and they'll be created or granted on an as-needed basis.
+Project-specific roles may be created and granted to Program members that are contributing to (or participating in) Program-related projects. The definition of these roles will depend on the project in question, and they'll be created or granted on an as-needed basis.
+
+These roles aren't detailed in this RFC, as they don't impact how the Program functions.
 
 ### Invitees
 
@@ -106,9 +108,13 @@ Once the poll has been closed, the vote should be interpreted as described by th
 
 While a poll is open and accepting votes, any Collaborator or Manager can use their right to veto by sending a message to the associated thread, detailing why they'd like to veto the vote.
 
-If another Collaborator or Manager isn't satisfied with the reasoning behind the veto, they can attempt to overrule the veto. To start this process, a new poll is created to allow voters to decide whether to overrule the veto. This poll lasts for three days and, if more than half of the individual votes are positive, the veto is overruled. If not, the original vote should be treated as if it has failed.
+If another Collaborator or Manager isn't satisfied with the reasoning behind the veto, they can attempt to overrule it. To start this process, a new poll is created to allow voters to decide whether to overrule the veto. This poll lasts for three days and, if more than half of the individual votes are positive, the veto is overruled. If this vote fails, the original vote should be treated as if it has failed.
 
-The veto process may be started on behalf of another Manager or Collaborator. Additionally, the list of voters on that veto may be kept secret. This is intended to avoid exposing those that supported the veto (or that didn't) to negative reactions or bullying by other Program members.
+As the veto process is operated at an individual level, Program members vote as individuals. Communities are not required to seek consensus and vote as a whole.
+
+If nobody attempts to overrule the veto before the original vote ends, the original vote should be treated as if it has failed. If someone attempts to overrule the veto, then the original vote will not be closed until the vote to overrule has ended, even if the vote to overrule finishes after the end date for the original vote.
+
+The veto process may be started on behalf of another Manager or Collaborator. Additionally, the list of voters on that veto may be kept secret. This is intended to avoid exposing those that started or supported the veto (or that didn't) to negative reactions or bullying by other Program members.
 
 **Note:** Individual Collaborators currently do not hold the right to veto.
 
@@ -137,6 +143,8 @@ The application process is defined as follows:
     * What the community rules are (or where to find them)
     * A full list of staff members and their roles
     * A method of accessing the community directly (for example, a Discord server invitation)
+    * A statement that shows that this RFC (and all other relevant public documentation) has been read and understood
+    * Whether the community intends to be part of the voting process and, if so, documentation on how that community will follow the voting process (or a commitment to figuring that out before the community may vote on anything)
 2. Once the above information has been provided, a Manager informs voting-eligible Program members on the Discord server, and opens a discussion thread to figure out how the Program members feel about the new community, and to discuss anything relevant to the application
 3. A Discord channel is set up to allow the administrators and moderators from the new community (who are to be invited and given the Invitee role) to talk with existing Program members regarding their application
 4. Once the voting-eligible Program members have informally agreed that enough information has been gathered, the vote may be created (with options to approve or deny the vote) --- it must be done in a channel the Invitees don't have access to, and follow the voting process outlined earlier in this document
--- a/structure/0062-quilt-community-collab.md
+++ b/structure/0062-quilt-community-collab.md
@@ -4,7 +4,7 @@ This document intends to describe the Quilt Community Collab Program, as well as
 
 # Motivation
 
-The Minecraft community is complex and diverse and, in order to keep everyone safe, it's often useful for communities to share information with each other. The main goal of the Quilt Community Collab Program is to facilitate this exchange of information between a selected group of communities, which have been vetted for trustworthiness and educated on the best way to approach this type of data.
+The Minecraft community is complex and diverse and, in order to keep everyone safe, it's often useful for communities to share information with each other. The main goal of the Quilt Community Collab Program is to facilitate this information-sharing between a selected group of communities, which have been vetted for trustworthiness and educated on the best way to approach this type of data.
 
 The Minecraft community at large has long been host to a surprisingly high amount of hatred and discrimination, and the Quilt Community Collab Program is one of our ways of giving back to the community that got us where we are --- with the aim of hopefully improving it for everyone.
 
@@ -66,7 +66,7 @@ Collaborators are given access to all moderation-related areas within Program sp
 
 The Individual Collaborator role is given to moderators that have joined the Program, but do not represent any specific community, which may be for any number of reasons. We believe that Individual Collaborators should be able to make use of the Program, and can contribute to it just how regular Collaborators can.
 
-Individuals are given the same level of access that regular Collaborators are given. However, Individual Collaborators are not given voting rights --- this is something that still needs to be discussed, but for the sake of simplicity and ease of implementation, this discussion has been deferred until later.
+Individual Collaborator are given the same level of access that regular Collaborators are given. Individual Collaborators are granted veto rights but, as the voting process happens at a community level, they are not currently given voting rights.
 
 ### Event Planners
 
@@ -177,7 +177,7 @@ Individuals that apply to the program are evaluated based on a number of criteri
 
 ## Process: Removing a community
 
-While we hope that it won't ever be necessary, it's important that we have the ability to remove communities from the program when it becomes necessary.
+While we hope that it won't ever be necessary, it's important that we have the ability to remove communities from the program if it should ever become necessary.
 
 ### Starting the process
 
@@ -206,7 +206,7 @@ Otherwise, **the vote has failed.** Internal discussions can be removed as neede
 
 ## Process: Removing an Individual Collaborator
 
-While we hope that it won't ever be necessary, it's important that we have the ability to remove Individual Collaborators from the program when it becomes necessary.
+While we hope that it won't ever be necessary, it's important that we have the ability to remove Individual Collaborators from the program if it should ever become necessary.
 
 This process matches the process for removing a community.
 
@@ -233,7 +233,7 @@ Several relevant votes were held on the Program's Discord server leading up to t
 4. Should we allow for external collaborators (now Individual Collaborators) that don't represent any specific community?
     - **21 voted:** Yes
     - **4 voted:** I don't care
-    - **0 voted:** Not
+    - **0 voted:** No
 
 # Prior art
 
--- a/structure/0062-quilt-community-collab.md
+++ b/structure/0062-quilt-community-collab.md
@@ -72,11 +72,11 @@ Once the poll is closed, the vote can be interpreted, depending on the specific 
 
 ## Veto rights
 
-While the vote is active, any Collaborator can use their veto right by sending a message inside the thread, detailing their reasoning.
+While the vote is active, any Collaborator or Manager can use their veto right by sending a message inside the thread, detailing their reasoning.
 
-If another Collaborator believes the reasoning isn’t satisfying, they can start the overruling process. If so, another poll is set up, if more than half of the individual votes are positive after three days, the veto is discarded.
+If another Collaborator or Manager believes the reasoning isn’t satisfying, they can start the overruling process. If so, another poll is set up, if more than half of the individual votes are positive after three days, the veto is discarded.
 
-If no Collaborator overrules the veto after three days, the vote is considered as failed right away. The Managers are free to only communicate the result to the Invitees at a later date to keep the veto secret.
+If no Collaborator or Manager overrules the veto after three days, the vote is considered as failed right away. The Managers are free to only communicate the result to the Invitees at a later date to keep the veto secret.
 
 ## Process: adding a community
 
--- a/structure/0062-quilt-community-collab.md
+++ b/structure/0062-quilt-community-collab.md
@@ -1,0 +1,125 @@
+# Summary
+
+This document describes the Quilt Community Collab program, the roles of each member and the associated processes.
+
+# Motivation
+
+The Minecraft community is a complex space, where it can be required for moderators to share intel between communities. The main goal of the Quilt Community Collab program is to facilitate exchange between a selected group of communities and improve user safety.
+
+The server also serves as a coordination server for events and other projects happening in the Minecraft sphere.
+
+# Explanations
+
+The Quilt Community Collab is a private space for Minecraft-centered communities to exchange information and coordinate. The main focus of the exchanges is set on moderation, but it can also be used for events and anything else that may be fitting.
+
+## What Community Collab isn’t
+
+- It isn’t an official affiliation with the Quilt project itself or any other community part of the program
+- It isn’t a place to blindly synchronise bans between communities
+    - Each community must decide on appropriate moderation action based on facts shown to them and their own rules and policies. Infringing this rule may get your community removed from the Community Collab program.
+
+## Roles inside the server
+
+### Keyholder
+
+The keyholder for this space is the same as for the Quilt Community Team. This role is described in the [RFC 0007 # Keyholders](https://github.com/QuiltMC/rfcs/blob/master/structure/0007-community-team.md#keyholders).
+
+### Managers
+
+Managers are responsible for keeping the collab program up and running.
+
+- Maintain and organise the associated Discord server
+- Actioning changes to the collab-related spaces
+- Organise polls and votes as required by formal processes
+- Help resolve conflicts between communities
+- Stay in touch with any community part of the program
+
+Community Collab Managers are the same as Community Managers of the Quilt Community Team.
+
+### Collaborators
+
+Collaborators are moderators of one or more communities in the program. They have access to the server to the moderation part of the server, such as general moderator chats and case threads. Access is also given to the event category.
+
+### Event Planners
+
+Event Planners are members of the staff team of one or more communities in the program, who focuses on event organisation. They have access to the event part of the server.
+
+### Project Roles
+
+Other roles may be given surrounding the different projects happening inside the program. Those roles are informal and will depend on each project.
+
+### Invitees
+
+Invitees are temporary users, with no access to the rest of the server. They only have access to temporary channels required by the process they are visiting for.
+
+## Voting process
+
+If required, a vote among the Community Collab group can be called by a Manager. A poll and an associated thread are created in an internal channel designated for that purpose, including as many details as possible.
+
+Each community gets a single vote representing its member. They are free to write their own process for determining their vote, but it must follow those rules:
+
+- This process must be democratic and give a voice to any community member who is able to join the Community Collab server under the Collaborator role.
+- Each eligible voter must have single, equal voting power.
+- The process can remove the voting power under some inactivity threshold.
+
+Once a community has decided their vote, one member can report their vote in the thread, mentioning the community name, and the option being voted for, with corresponding proof (such as a screenshot of an internal poll).
+
+Each community also has the option of skipping its vote if none of its members has an opinion on the poll. Such instances should be reported the same way.
+
+Polls must include a closing date of at least ten days. If required, the closing date can be pushed away by a maximum of a month, upon the informal agreement with the majority of the other communities. During that period, any Collaborator can use their veto right (see below).
+
+Once the poll is closed, the vote can be interpreted, depending on the specific process this vote is part of, if any.
+
+## Veto rights
+
+While the vote is active, any Collaborator can use their veto right by sending a message inside the thread, detailing their reasoning.
+
+If another Collaborator believes the reasoning isn’t satisfying, they can start the overruling process. If so, another poll is set up, if more than half of the individual votes are positive after three days, the veto is discarded.
+
+If no Collaborator overrules the veto after three days, the vote is considered as failed right away. The Managers are free to only communicate the result to the Invitees at a later date to keep the veto secret.
+
+## Process: adding a community
+
+Any community, ideally having their main user space on Discord, centred around Minecraft, can apply to join the Community Collab program. The process is defined as follows:
+
+1. A member of the community who wishes to join, ideally someone holding administrative and moderation powers, should get in touch with a Manager to start the process. The following questions should be asked to the community and collected, along with an invite to the server:
+    - What is the main focus of the server?
+    - Why is your community looking to join the Community Collab program?
+    - What is your current ruleset?
+    - Who is currently part of your staff team, and who would have access to Community Collab?
+2. Once those answers are collected, the Manager can inform all the Collaborators about the community, and start a discussion to gauge the interest of the community.
+3. The members of the community are invited to the Community Collab server under the Invitee role and a channel is set up to allow discussion between Collaborators and Invitee.
+4. Once the Collaborators informally agree that enough information has been gathered, the vote can be created with the option to approve or deny, according to the process defined in this document. The Invitees should be informed the vote has begun.
+5. After the poll is closed, the vote is considered passed if more than 75% of the votes cast are positive. The discussion channel must be archived and kept read-only.
+- If the vote has passed: the Invitees are elevated to the rank of Collaborators. Internal discussions can be purged first if needs be. The community is added to any listing that may exist.
+- If the vote is rejected: enough time is given to the Invitees to acknowledge the result and leave the server. Managers are free to kick Invitees if necessary.
+
+# Rationale and alternatives
+
+The Quilt Community Collab program existed prior to this document. This document tries to summarize how the server has been working so far while formalising some aspects that were still left up in the air.
+
+More importantly, three polls have been held inside the server, with the following results:
+
+- When voting for changes on who is or isn't part of the CC network, how should we count votes?
+    - 10 answered: One vote per community: each community will have to write down a formal process on which they decide to vote for or against changes. This process must be democratic and fair.
+    - 4 answered: One vote per collaborator: each user or plural system present in the server gets one vote
+- When voting on who is or isn't part of the CC network, what proportions of approvals versus neutral votes are required to pass? (multiple answers accepted)
+    - 11 answered: >75% approval to pass (there must be at least 3 times more approval than neutral)
+    - 4 answered: >50% approval to pass (there must be more approval than neutral)
+    - 1 answered: 100% approval to pass (there are no neutral votes, only vetos)
+- When voting for changes on who is or isn't part of the CC network, should we have veto rights, and if so how should they work?
+    - 17 answered: Any voter can veto, which can then be overruled by the majority if the concerns aren't believed to be deserved
+    - Nobody answered: There are no veto rights.
+    - Nobody answered: Any voter can veto, which will immediately cancel the vote
+
+# Prior art
+
+Quilt Community Collab is an indirect replacement of the Modding Interpol, with a wider range of activities, such as events and related projects
+
+# Unresolved questions
+
+The following questions will have to be solved at a later date:
+
+- This currently doesn’t define how to remove communities
+- We aren’t defining any role for administrators of the communities part of the program
+- We only allow for full communities to join, we should look at whenever we also want to invite individuals
--- a/structure/0062-quilt-community-collab.md
+++ b/structure/0062-quilt-community-collab.md
@@ -15,7 +15,7 @@ The Quilt Community Collab is a private space for Minecraft-centered communities
 ## What Community Collab isn’t
 
 - It isn’t an official affiliation with the Quilt project itself or any other community part of the program
-- It isn’t a place to blindly synchronise bans between communities
+- It isn’t a place to blindly synchronise punishments between communities
     - Each community must decide on appropriate moderation action based on facts shown to them and their own rules and policies. Infringing this rule may get your community removed from the Community Collab program.
 
 ## Roles inside the server
@@ -42,7 +42,7 @@ Collaborators are moderators of one or more communities in the program. They hav
 
 ### Event Planners
 
-Event Planners are members of the staff team of one or more communities in the program, who focuses on event organisation. They have access to the event part of the server.
+Event Planners are members of the staff team of one or more communities in the program, who focus on event organisation. They have access to the events part of the server.
 
 ### Project Roles
 
@@ -54,13 +54,13 @@ Invitees are temporary users, with no access to the rest of the server. They onl
 
 ## Voting process
 
-If required, a vote among the Community Collab group can be called by a Manager. A poll and an associated thread are created in an internal channel designated for that purpose, including as many details as possible.
+If required, a vote can be called by a Manager where the involved Collab communities are each constituents. A poll and an associated thread are created in an internal channel designated for that purpose, including as many details as possible.
 
-Each community gets a single vote representing its member. They are free to write their own process for determining their vote, but it must follow those rules:
+Each community gets a single vote. They are free to write their own process for determining their vote, but it must follow those rules:
 
 - This process must be democratic and give a voice to any community member who is able to join the Community Collab server under the Collaborator role.
 - Each eligible voter must have single, equal voting power.
-- The process can remove the voting power under some inactivity threshold.
+- The process can remove the voting power under some reasonable inactivity threshold.
 
 Once a community has decided their vote, one member can report their vote in the thread, mentioning the community name, and the option being voted for, with corresponding proof (such as a screenshot of an internal poll).
 
--- a/structure/0062-quilt-community-collab.md
+++ b/structure/0062-quilt-community-collab.md
@@ -1,135 +1,201 @@
 # Summary
 
-This document describes the Quilt Community Collab program, the roles of each member and the associated processes.
+This document intends to describe the Quilt Community Collab Program, as well as its entry requirements, processes, and the roles that may be assigned to its members.
 
 # Motivation
 
-The Minecraft community is a complex space, where it can be required for moderators to share intel between communities. The main goal of the Quilt Community Collab program is to facilitate exchange between a selected group of communities and improve user safety.
+The Minecraft community is complex and diverse and, in order to keep everyone safe, it's often useful for communities to share information with each other. The main goal of the Quilt Community Collab Program is to facilitate this exchange of information between a selected group of communities, which have been vetted for trustworthiness and educated on the best way to approach this type of data.
 
-The server also serves as a coordination server for events and other projects happening in the Minecraft sphere.
+The Minecraft community at large has long been host to a surprisingly high amount of hatred and discrimination, and the Quilt Community Collab Program is one of our ways of giving back to the community that got us where we are --- with the aim of hopefully improving it for everyone.
 
-# Explanations
+# Explanation
 
-The Quilt Community Collab is a private space for Minecraft-centered communities to exchange information and coordinate. The main focus of the exchanges is set on moderation, but it can also be used for events and anything else that may be fitting.
+The Quilt Community Collab Program (henceforth, "the Program") largely exists as a private Discord server, wherein representatives from participating Minecraft-related communities exchange information and coordinate on issues of moderation, alongside Quilt's Community Team and individual contributors that don't represent specific communities.
 
-## What Community Collab isn’t
+It also exists to serve as a coordination point for events and other cross-community projects within the Minecraft sphere.
 
-- It isn’t an official affiliation with the Quilt project itself or any other community part of the program
-- It isn’t a place to blindly synchronise punishments between communities
-    - Each community must decide on appropriate moderation action based on facts shown to them and their own rules and policies. Infringing this rule may get your community removed from the Community Collab program.
+## What the Program is not
 
-## Roles inside the server
+* An official, direct affiliation with the Quilt project, or any other community that participates in the Program
+* A space for building or coordinating the character assassinations of community members or prominent personalities
+* A space meant for blindly synchronizing punishments between communities
+  * Instead, each community must decide on appropriate moderation actions for their own spaces, which may be informed by the data they're given access to, but not blindly copied between communities --- Breaking this rule is likely to result in a community being removed from the Program
+
+## Program Roles
+
+As the Program exists largely on Discord, the roles of its members are directly represented by Discord server roles.
 
 ### Keyholder
 
-The keyholder for this space is the same as for the Quilt Community Team. This role is described in the [RFC 0007 # Keyholders](https://github.com/QuiltMC/rfcs/blob/master/structure/0007-community-team.md#keyholders).
+The Keyholder for this space is the same as for the other Quilt spaces. This role is described in the Keyholders section of [RFC 7: Community Team](https://github.com/QuiltMC/rfcs/blob/master/structure/0007-community-team.md#keyholders), and works in the same way.
 
 ### Managers
 
-Managers are responsible for keeping the collab program up and running.
+Managers are responsible for keeping the Program up-and-running. They take care of the following tasks, among other things:
 
-- Maintain and organise the associated Discord server
-- Actioning changes to the collab-related spaces
-- Organise polls and votes as required by formal processes
-- Help resolve conflicts between communities
-- Stay in touch with any community part of the program
+* Maintaining, organizing, and actioning changes to the Program's Discord server, and any other Program-related spaces
+* Organizing polls and votes, as required by formal procedures
+* Aiding in conflict resolution between communities and Program members
+* Staying in touch (and maintaining relations) with the communities that are part of the Program
+* Procuring and maintaining any services that the Program needs to function properly
 
-Community Collab Managers are the same as Community Managers of the Quilt Community Team.
+This role is very similar to the Community Manager role within Quilt's own Community Team.
 
 ### Collaborators
 
-Collaborators are moderators of one or more communities in the program. They have access to the server to the moderation part of the server, such as general moderator chats and case threads. Access is also given to the event category.
+The Collaborator role is given to moderators from the communities that are part of the Program, and who have joined a Program-related space in order to represent their communities.
+
+Collaborators are given access to all moderation-related areas within Program spaces, such as general moderation chat channels and case-tracking threads. Additionally, all Collaborators are granted access to spaces dedicated to organizing events between participating communities.
 
 ### Event Planners
 
-Event Planners are members of the staff team of one or more communities in the program, who focus on event organisation. They have access to the events part of the server.
+The Event Planner role is given to members of participating communities' community teams that focus on event creation, execution and management. They do not have access to moderation-related areas within Program spaces, and are instead given their own dedicated spaces.
 
 ### Project Roles
 
-Other roles may be given surrounding the different projects happening inside the program. Those roles are informal and will depend on each project.
+Informal roles may be created and granted to Program members that are contributing to (or participating in) Program-related projects. The definition of these roles will depend on the project in question, and they'll be created or granted on an as-needed basis.
 
 ### Invitees
 
-Invitees are temporary users, with no access to the rest of the server. They only have access to temporary channels required by the process they are visiting for.
+Invitees are temporary users with no access to Program-related spaces, other than temporary areas that are created for whatever purpose they're visiting the respective space.
 
 ## Voting process
 
-If required, a vote can be called by a Manager where the involved Collab communities are each constituents. A poll and an associated thread are created in an internal channel designated for that purpose, including as many details as possible.
+Votes may be called by Program Managers, where each of the communities participating in the Program represent the poll's voters. As the Program exists largely on Discord, a poll and an associated thread are created on the Program's Discord server, in a channel designated for this purpose.
 
-Each community gets a single vote. They are free to write their own process for determining their vote, but it must follow those rules:
+The poll should be as detailed as possible in order to help everyone to make an informed decision. Discussions related to the poll should happen in the attached thread, to try to keep everything together.
 
-- This process must be democratic and give a voice to any community member who is able to join the Community Collab server under the Collaborator role.
-- Each eligible voter must have single, equal voting power.
-- The process can remove the voting power under some reasonable inactivity threshold.
+Each community is assigned one single vote. Individual communities are responsible for coming up with their own process for determining their vote, but it must follow these rules:
 
-Once a community has decided their vote, one member can report their vote in the thread, mentioning the community name, and the option being voted for, with corresponding proof (such as a screenshot of an internal poll).
+* The process must be democratic and give a voice to any community member that is otherwise eligible to join the Program's Discord server under the Collaborator role, even if they're not currently present on the server
+* Each eligible member of the community must have equal voting power
+* Eligible members may only be excluded via a reasonable inactivity threshold
+* The process must be formally documented somewhere, but this documentation may be kept internal to the given community
+* Participating communities must allow the Program's Manager team to audit their process, which may be motivated by reports from the given community's members, or from other Program members, including the Managers themselves
 
-Each community also has the option of skipping its vote if none of its members has an opinion on the poll. Such instances should be reported the same way.
+Once a community has decided how it's going to vote, it must designate a single eligible community member to represent that vote. This representative must report their community's vote in the attached thread, specifically mentioning the community's name and the option being voted for, and providing proof of the results --- such as a screenshot of an internal poll.
 
-Polls must include a closing date of at least ten days. If required, the closing date can be pushed away by a maximum of a month, upon the informal agreement with the majority of the other communities. During that period, any Collaborator can use their veto right (see below), if allowed by the process.
+Participating communities also have the option of abstaining from the vote. This must either be reported in the thread as described above or, in the case of a community that wishes to abstain from all votes, directly to a Program Manager in advance. Regardless of the method, documented consensus among eligible community members must be provided.
 
-Once the poll is closed, the vote can be interpreted, depending on the specific process this vote is part of, if any.
+All polls must include a closing date that is set at least ten days into the future. If needed, this date may be pushed back by up to a month, as long as an informal agreement between the majority of the Program's participating, non-abstaining communities has been reached, and documented within the thread attached to the poll.
 
-## Veto rights
+At any point before the poll has closed, any Collaborator can use their right to veto, if allowed by the process. This right is described below.
 
-While the vote is active, any Collaborator or Manager can use their veto right by sending a message inside the thread, detailing their reasoning.
+Once the poll has been closed, the vote should be interpreted as described by the process the poll is a part of.
 
-If another Collaborator or Manager believes the reasoning isn’t satisfying, they can start the overruling process. If so, another poll is set up, if more than half of the individual votes are positive after three days, the veto is discarded.
+### The right to veto
 
-If no Collaborator or Manager overrules the veto after three days, the vote is considered as failed right away. The Managers are free to only communicate the result to the Invitees at a later date to keep the veto secret.
+While a poll is open and accepting votes, any Collaborator or Manager can use their right to veto by sending a message to the attached thread, detailing why they'd like to veto the vote.
 
-## Process: adding a community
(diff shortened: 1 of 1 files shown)


structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+Other roles may be given surrounding the different projects happening inside the program. Those roles are informal and will depend on each project.
+
+### Invitees
+
+Invitees are temporary users, with no access to the rest of the server. They only have access to temporary channels required by the process they are visiting for.
+
+## Voting process
+
+If required, a vote among the Community Collab group can be called by a Manager. A poll and an associated thread are created in an internal channel designated for that purpose, including as many details as possible.
```
wafflecoffee, line 57:
```suggestion
If required, a vote can be called by a Manager where the involved Collab communities are each constituents. A poll and an associated thread are created in an internal channel designated for that purpose, including as many details as possible.
```

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+Each community gets a single vote representing its member. They are free to write their own process for determining their vote, but it must follow those rules:
+
+- This process must be democratic and give a voice to any community member who is able to join the Community Collab server under the Collaborator role.
+- Each eligible voter must have single, equal voting power.
+- The process can remove the voting power under some inactivity threshold.
+
+Once a community has decided their vote, one member can report their vote in the thread, mentioning the community name, and the option being voted for, with corresponding proof (such as a screenshot of an internal poll).
```
wafflecoffee, line 65: I feel as if this is asking too much of a lot of communities. If I had to guess, very few servers actually have any sort of formal voting system - Quilt is the first place I've been which actually has one.

Additionally, some communities are considerably more active in CC than others, and there are some that I haven't seen participating at all. What happens if simply nobody from the community responds, ever? I don't think it would be reasonable for us to go after them for this,

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+Invitees are temporary users, with no access to the rest of the server. They only have access to temporary channels required by the process they are visiting for.
+
+## Voting process
+
+If required, a vote among the Community Collab group can be called by a Manager. A poll and an associated thread are created in an internal channel designated for that purpose, including as many details as possible.
+
+Each community gets a single vote representing its member. They are free to write their own process for determining their vote, but it must follow those rules:
+
+- This process must be democratic and give a voice to any community member who is able to join the Community Collab server under the Collaborator role.
+- Each eligible voter must have single, equal voting power.
```
wafflecoffee, line 62: Quilt's own CT voting process violates this rule! Admin Board gets three votes if there's a tie.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+Managers are responsible for keeping the collab program up and running.
+
+- Maintain and organise the associated Discord server
+- Actioning changes to the collab-related spaces
+- Organise polls and votes as required by formal processes
+- Help resolve conflicts between communities
+- Stay in touch with any community part of the program
+
+Community Collab Managers are the same as Community Managers of the Quilt Community Team.
```
SilverAndro, line 37: Why is this a requirement if its not officially affiliated with quilt? Wouldn't it make more sense to declare that current CMs are managers but future ones can be elected or something?

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+- This process must be democratic and give a voice to any community member who is able to join the Community Collab server under the Collaborator role.
+- Each eligible voter must have single, equal voting power.
+- The process can remove the voting power under some inactivity threshold.
+
+Once a community has decided their vote, one member can report their vote in the thread, mentioning the community name, and the option being voted for, with corresponding proof (such as a screenshot of an internal poll).
+
+Each community also has the option of skipping its vote if none of its members has an opinion on the poll. Such instances should be reported the same way.
+
+Polls must include a closing date of at least ten days. If required, the closing date can be pushed away by a maximum of a month, upon the informal agreement with the majority of the other communities. During that period, any Collaborator can use their veto right (see below).
```
wafflecoffee, line 69: Overall this process seems way too lengthy and involved, especially for things that in the end might have barely any effect, such as adding another community to the program. We already know what happened with how complex the CT voting system is, and this is... even worse, imo

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+Once the poll is closed, the vote can be interpreted, depending on the specific process this vote is part of, if any.
+
+## Veto rights
+
+While the vote is active, any Collaborator can use their veto right by sending a message inside the thread, detailing their reasoning.
+
+If another Collaborator believes the reasoning isn’t satisfying, they can start the overruling process. If so, another poll is set up, if more than half of the individual votes are positive after three days, the veto is discarded.
+
+If no Collaborator overrules the veto after three days, the vote is considered as failed right away. The Managers are free to only communicate the result to the Invitees at a later date to keep the veto secret.
```
wafflecoffee, line 79: This seems wayyyyy too easily abusable.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+Community Collab Managers are the same as Community Managers of the Quilt Community Team.
+
+### Collaborators
+
+Collaborators are moderators of one or more communities in the program. They have access to the server to the moderation part of the server, such as general moderator chats and case threads. Access is also given to the event category.
+
+### Event Planners
+
+Event Planners are members of the staff team of one or more communities in the program, who focuses on event organisation. They have access to the event part of the server.
```
SilverAndro, line 45:
```suggestion
Event Planners are members of the staff team of one or more communities in the program, who focus on event organisation. They have access to the events part of the server.
```

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+Polls must include a closing date of at least ten days. If required, the closing date can be pushed away by a maximum of a month, upon the informal agreement with the majority of the other communities. During that period, any Collaborator can use their veto right (see below).
+
+Once the poll is closed, the vote can be interpreted, depending on the specific process this vote is part of, if any.
+
+## Veto rights
+
+While the vote is active, any Collaborator can use their veto right by sending a message inside the thread, detailing their reasoning.
+
+If another Collaborator believes the reasoning isn’t satisfying, they can start the overruling process. If so, another poll is set up, if more than half of the individual votes are positive after three days, the veto is discarded.
```
wafflecoffee, line 77: Is a Manager not allowed to overrule a veto?

Setting up a whole other poll for this will get *real* confusing, *real* fast.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+Event Planners are members of the staff team of one or more communities in the program, who focuses on event organisation. They have access to the event part of the server.
+
+### Project Roles
+
+Other roles may be given surrounding the different projects happening inside the program. Those roles are informal and will depend on each project.
+
+### Invitees
+
+Invitees are temporary users, with no access to the rest of the server. They only have access to temporary channels required by the process they are visiting for.
```
SilverAndro, line 53: I feel like this should probably have a better name? When I read this at first I thought you were talking about how new joins need to get roles assigned

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+    - 4 answered: >50% approval to pass (there must be more approval than neutral)
+    - 1 answered: 100% approval to pass (there are no neutral votes, only vetos)
+- When voting for changes on who is or isn't part of the CC network, should we have veto rights, and if so how should they work?
+    - 17 answered: Any voter can veto, which can then be overruled by the majority if the concerns aren't believed to be deserved
+    - Nobody answered: There are no veto rights.
+    - Nobody answered: Any voter can veto, which will immediately cancel the vote
+
+# Prior art
+
+Quilt Community Collab is an indirect replacement of the Modding Interpol, with a wider range of activities, such as events and related projects
```
wafflecoffee, line 117: Needs a full stop

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Invitees
+
+Invitees are temporary users, with no access to the rest of the server. They only have access to temporary channels required by the process they are visiting for.
+
+## Voting process
+
+If required, a vote among the Community Collab group can be called by a Manager. A poll and an associated thread are created in an internal channel designated for that purpose, including as many details as possible.
+
+Each community gets a single vote representing its member. They are free to write their own process for determining their vote, but it must follow those rules:
```
SilverAndro, line 59:
```suggestion
Each community gets a single vote. They are free to write their own process for determining their vote, but it must follow those rules:
```

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+# Prior art
+
+Quilt Community Collab is an indirect replacement of the Modding Interpol, with a wider range of activities, such as events and related projects
+
+# Unresolved questions
+
+The following questions will have to be solved at a later date:
+
+- This currently doesn’t define how to remove communities
```
wafflecoffee, line 123: This should absolutely be addressed in this document. There is already prior art as we have removed at least one community before. I think it would be easiest to adopt/adapt the current Quilt team removal process.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+The server also serves as a coordination server for events and other projects happening in the Minecraft sphere.
+
+# Explanations
+
+The Quilt Community Collab is a private space for Minecraft-centered communities to exchange information and coordinate. The main focus of the exchanges is set on moderation, but it can also be used for events and anything else that may be fitting.
+
+## What Community Collab isn’t
+
+- It isn’t an official affiliation with the Quilt project itself or any other community part of the program
+- It isn’t a place to blindly synchronise bans between communities
```
wafflecoffee, line 18:
```suggestion
- It isn’t a place to blindly synchronise punishments between communities
```

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Voting process
+
+If required, a vote among the Community Collab group can be called by a Manager. A poll and an associated thread are created in an internal channel designated for that purpose, including as many details as possible.
+
+Each community gets a single vote representing its member. They are free to write their own process for determining their vote, but it must follow those rules:
+
+- This process must be democratic and give a voice to any community member who is able to join the Community Collab server under the Collaborator role.
+- Each eligible voter must have single, equal voting power.
+- The process can remove the voting power under some inactivity threshold.
```
SilverAndro, line 63: Minor loophole thats probably not much of an issue but theoretically you could make the inactivity threshold to something like 12 hours? No harm in being specific at least
```suggestion
- The process can remove the voting power under some reasonable inactivity threshold.
```

wafflecoffee (changes requested):

structure/0062-quilt-community-collab.md
SilverAndro, line 79: Why are we using the generic term for anyone with temporary access here? I think the definition of "Invitee" might need to be pulled in to better represent what youre envisioning, or declare a different term altogether

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+    - 10 answered: One vote per community: each community will have to write down a formal process on which they decide to vote for or against changes. This process must be democratic and fair.
+    - 4 answered: One vote per collaborator: each user or plural system present in the server gets one vote
+- When voting on who is or isn't part of the CC network, what proportions of approvals versus neutral votes are required to pass? (multiple answers accepted)
+    - 11 answered: >75% approval to pass (there must be at least 3 times more approval than neutral)
+    - 4 answered: >50% approval to pass (there must be more approval than neutral)
+    - 1 answered: 100% approval to pass (there are no neutral votes, only vetos)
+- When voting for changes on who is or isn't part of the CC network, should we have veto rights, and if so how should they work?
+    - 17 answered: Any voter can veto, which can then be overruled by the majority if the concerns aren't believed to be deserved
+    - Nobody answered: There are no veto rights.
+    - Nobody answered: Any voter can veto, which will immediately cancel the vote
```
SilverAndro, line 113: I feel like this section should be near the top (and rewritten a bit to make more sense with that ordering)

structure/0062-quilt-community-collab.md
SilverAndro, line 123: This seems kind of important though? I'd really prefer to get this down before we start pulling misc. communities in

SilverAndro (changes requested):

structure/0062-quilt-community-collab.md
Akarys42: Not since #53 has been merged

structure/0062-quilt-community-collab.md
Akarys42: The voting process itself takes 10 days under normal circumstances. We need to provide each community with a decently long time to get a poll set up, vote and report the result. This is only three days longer than the CT process, because of that overhead.

structure/0062-quilt-community-collab.md
Akarys42: That's not a bad point, although I think we should defer that to a later date. I can add it to the list of unresolved questions if that works for you.

structure/0062-quilt-community-collab.md
Akarys42: Any idea of what a better name would be?

structure/0062-quilt-community-collab.md
Akarys42: If a community doesn't report anything, they simply aren't counted in the poll, which is fine. I would be okay with a message such as "does everyone agree to X" with a few upvotes on them. We at least need something for transparency.

structure/0062-quilt-community-collab.md
Akarys42: Veto has been given to managers in dc8021b. I would like to avoid the double poll issue, any suggestions?

structure/0062-quilt-community-collab.md
Akarys42: One solution I would have is to allow an overruling up to the end of the poll, with a minimum of 3 days

structure/0062-quilt-community-collab.md
Akarys42: I disagree, this is part of the rationale. The information someone might be looking for should be at the top of the document, the rationale at the end because that's a section you rarely read.

structure/0062-quilt-community-collab.md
SilverAndro: Sure makes sense

structure/0062-quilt-community-collab.md
SilverAndro: I suppose, it just makes more sense to me to flow in `What -> Why -> How` than `What -> How -> Why` but thats me

Akarys42: It is the same ordering as the template, to be fair

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+Each community also has the option of skipping its vote if none of its members has an opinion on the poll. Such instances should be reported the same way.
+
+Polls must include a closing date of at least ten days. If required, the closing date can be pushed away by a maximum of a month, upon the informal agreement with the majority of the other communities. During that period, any Collaborator can use their veto right (see below).
+
+Once the poll is closed, the vote can be interpreted, depending on the specific process this vote is part of, if any.
+
+## Veto rights
+
+While the vote is active, any Collaborator or Manager can use their veto right by sending a message inside the thread, detailing their reasoning.
```
sciwhiz12, line 75: What happens if a Collaborator/Manager withdraws their veto? 
Can a Collaborator/Manager make multiple vetos based on different reasonings (one after another)?

My opinion is that a veto should require a detailed explanation and reasoning to be valid, and that a veto which does not provide such explanation/reasoning to be held as invalid.

Also, to clarify, each community gets one equal vote, but each Collaborator (of which a community may have many) gets one veto?

structure/0062-quilt-community-collab.md
SilverAndro: Well the template is wrong then 😄

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+Each community gets a single vote. They are free to write their own process for determining their vote, but it must follow those rules:
+
+- This process must be democratic and give a voice to any community member who is able to join the Community Collab server under the Collaborator role.
+- Each eligible voter must have single, equal voting power.
+- The process can remove the voting power under some reasonable inactivity threshold.
+
+Once a community has decided their vote, one member can report their vote in the thread, mentioning the community name, and the option being voted for, with corresponding proof (such as a screenshot of an internal poll).
+
+Each community also has the option of skipping its vote if none of its members has an opinion on the poll. Such instances should be reported the same way.
```
sciwhiz12, line 67: To clarify, is this the method of abstaining from a vote? 

What happens if a community does not vote? Is that counted in the same way as a skipped/abstained vote under this clause?

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+If another Collaborator or Manager believes the reasoning isn’t satisfying, they can start the overruling process. If so, another poll is set up, if more than half of the individual votes are positive after three days, the veto is discarded.
+
+If no Collaborator or Manager overrules the veto after three days, the vote is considered as failed right away. The Managers are free to only communicate the result to the Invitees at a later date to keep the veto secret.
+
+## Process: adding a community
+
+Any community, ideally having their main user space on Discord, centred around Minecraft, can apply to join the Community Collab program. The process is defined as follows:
+
+1. A member of the community who wishes to join, ideally someone holding administrative and moderation powers, should get in touch with a Manager to start the process. The following questions should be asked to the community and collected, along with an invite to the server:
```
sciwhiz12, line 85: It should be required that a member of the community that wishes to join must be someone with administrative (or moderation) powers in that (Discord) community and with active moderation duties. 

This will mean that the Invitees of that community will be members who can represent the community by themselves (via their administrative/moderation powers), rather than possibly impotent members who can only advise the acutal administration or moderation team of the community (middlemen).

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+    - What is the main focus of the server?
+    - Why is your community looking to join the Community Collab program?
+    - What is your current ruleset?
+    - Who is currently part of your staff team, and who would have access to Community Collab?
+2. Once those answers are collected, the Manager can inform all the Collaborators about the community, and start a discussion to gauge the interest of the community.
+3. The members of the community are invited to the Community Collab server under the Invitee role and a channel is set up to allow discussion between Collaborators and Invitee.
+4. Once the Collaborators informally agree that enough information has been gathered, the vote can be created with the option to approve or deny, according to the process defined in this document. The Invitees should be informed the vote has begun.
```
sciwhiz12, line 92: Do Invitees have access to see the vote and comment on its ongoing result? It should be clarified in the document.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+    - 4 answered: >50% approval to pass (there must be more approval than neutral)
+    - 1 answered: 100% approval to pass (there are no neutral votes, only vetos)
+- When voting for changes on who is or isn't part of the CC network, should we have veto rights, and if so how should they work?
+    - 17 answered: Any voter can veto, which can then be overruled by the majority if the concerns aren't believed to be deserved
+    - Nobody answered: There are no veto rights.
+    - Nobody answered: Any voter can veto, which will immediately cancel the vote
+
+# Prior art
+
+Quilt Community Collab is an indirect replacement of the Modding Interpol, with a wider range of activities, such as events and related projects.
```
sciwhiz12, line 117: Could you explain in a bit more detail what the "Modding Interpol" is?

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+
+# Unresolved questions
+
+The following questions will have to be solved at a later date:
+
+- This currently doesn’t define how to remove communities
+- We aren’t defining any role for administrators of the communities part of the program
+- We only allow for full communities to join, we should look at whenever we also want to invite individuals
```
sciwhiz12, line 125: What examples would we have of individuals being invited to the program in their personal capacity (rather than in a representative capacity for a community)?

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,125 @@
...
+# Prior art
+
+Quilt Community Collab is an indirect replacement of the Modding Interpol, with a wider range of activities, such as events and related projects.
+
+# Unresolved questions
+
+The following questions will have to be solved at a later date:
+
+- This currently doesn’t define how to remove communities
+- We aren’t defining any role for administrators of the communities part of the program
```
sciwhiz12, line 124: One could argue that an administrator of a community ought to be that which initiates the process of adding a community to the program. They can then designate other members of their staff, such as moderators, to be added in their stead.

sciwhiz12 (changes requested):

structure/0062-quilt-community-collab.md
Akarys42: > What happens if a Collaborator/Manager withdraws their veto?

The vote is considered failed right away, and we carry on. The same goes for any other vote, I don't think there is any point in formalizing all of those points.

> Can a Collaborator/Manager make multiple vetos based on different reasonings (one after another)?

I guess not, they should be combined together

> My opinion is that a veto should require a detailed explanation and reasoning to be valid, and that a veto which does not provide such explanation/reasoning to be held as invalid.

This will be up to the other Collaborators to enforce through overruling. I don't want to have a single person be able to decide that.

> Also, to clarify, each community gets one equal vote, but each Collaborator (of which a community may have many) gets one veto?

To simply things, yes. It doesn't change anything, to be fair.

structure/0062-quilt-community-collab.md
Akarys42: Yes, this is how a community can abstain.

The interpretation of a community not responding will depend on each specific process. For adding new communities, the threshold is defined depending on the number of communities who cast a vote and not all the eligible communities, meaning in that case, abstaining and not responding are the same thing.

structure/0062-quilt-community-collab.md
Akarys42: At the moment we do not, but we might want to add some individual members, which is why this is a future concern

structure/0062-quilt-community-collab.md
Akarys42: That message refers to the fact they currently aren't able to be in the server, and we may want to think about a better solution on the long term

structure/0062-quilt-community-collab.md
Akarys42: Clarified in 57c6818

structure/0062-quilt-community-collab.md
Akarys42: Clarified in 57c6818

structure/0062-quilt-community-collab.md
Akarys42: I've added a few more words in 4e73df6. That is about all I know.

structure/0062-quilt-community-collab.md
Akarys42: Added in e4357f8

structure/0062-quilt-community-collab.md
sciwhiz12: > The vote is considered failed right away, and we carry on. The same goes for any other vote, I don't think there is any point in formalizing all of those points.

Wait, the _vote_ is considered failed when a veto is withdrawn by the party which used it?

> This will be up to the other Collaborators to enforce through overruling. I don't want to have a single person be able to decide that.

> To simply things, yes. It doesn't change anything, to be fair.

One could argue that a community could gum up the works of a vote by having their multiple Collaborators cast their veto with any reason one after another, given that we neither enforce a limit on the number of Collaborators from a community, nor require a reason for casting a veto (which puts the burden of overruling reasonless vetoes on other Collaborators/Managers).

Then again, if a community does try this obstructionist tactic, they may find themselves at the receiving end of a removal vote afterwards. 🤔

Also, what I meant by providing an explanation/reasoning is any comprehensible reason beyond "I cast my veto", so a veto _must_ have some sort of public reasoning on why it was cast. I draw inspiration from the [OpenJDK Bylaws' part on a Veto in the section for Consensus Voting](https://openjdk.org/bylaws#veto):

> A **Veto** must be accompanied by a justification; a Veto without a justification is invalid. If a Veto is raised then the voter who raised the veto and those who support the proposed action must work together to reach a mutually-agreeable resolution. If a Veto is resolved then the voter who raised the Veto must withdraw that vote by explicitly abstaining or by casting a Yes vote.

---

Also, with regards to a Collaborator/Manager's one veto, does that mean once they cast their veto, regardless of its resolution (withdrawn, overruled), they cannot cast another veto in the same vote? Are they allowed to use their veto right multiple times in a single vote?

Akarys42:
A process for removing communities has been added in 0f2e81c

structure/0062-quilt-community-collab.md
Akarys42: > Wait, the vote is considered failed when a veto is withdrawn by the party which used it?

Sorry, that wasn't clear. If withdrawn, the veto is considered as overruled immediately, if we really need to tie it to the current process, or pretend it didn't happen, that's the same result.

> Then again, if a community does try this obstructionist tactic, they may find themselves at the receiving end of a removal vote afterwards. 

Exactly, we assume good faith from communities here. Acting in bad faith would probably mean a vote against them soon enough. 

> Also, with regards to a Collaborator/Manager's one veto, does that mean once they cast their veto, regardless of its resolution (withdrawn, overruled), they cannot cast another veto in the same vote? Are they allowed to use their veto right multiple times in a single vote?

There are no specific rules on that. If the person has two concerns, they would be better off combining them into a single veto to have better chances of a common agreement. I don't want to enforce having only a single veto, though, in case other elements come in later during the voting process and do need another veto. I think we can also assume good faith here.

SilverAndro (approved):

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
+# Summary
+
+This document describes the Quilt Community Collab program, the roles of each member and the associated processes.
+
+# Motivation
+
+The Minecraft community is a complex space, where it can be required for moderators to share intel between communities. The main goal of the Quilt Community Collab program is to facilitate exchange between a selected group of communities and improve user safety.
```
gdude2002, line 7: I'd probably say useful, not required, here

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+
+The server also serves as a coordination server for events and other projects happening in the Minecraft sphere.
+
+# Explanations
+
+The Quilt Community Collab is a private space for Minecraft-centered communities to exchange information and coordinate. The main focus of the exchanges is set on moderation, but it can also be used for events and anything else that may be fitting.
```
gdude2002, line 13: It's The Quilt Community Collab Program, it doesn't really make sense in English without that qualifier

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+### Keyholder
+
+The keyholder for this space is the same as for the Quilt Community Team. This role is described in the [RFC 0007 # Keyholders](https://github.com/QuiltMC/rfcs/blob/master/structure/0007-community-team.md#keyholders).
+
+### Managers
+
+Managers are responsible for keeping the collab program up and running.
+
+- Maintain and organise the associated Discord server
+- Actioning changes to the collab-related spaces
```
gdude2002, line 32: The tense of this point doesn't match the others, though I think this tense would work better for all of the points

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+
+The server also serves as a coordination server for events and other projects happening in the Minecraft sphere.
+
+# Explanations
+
+The Quilt Community Collab is a private space for Minecraft-centered communities to exchange information and coordinate. The main focus of the exchanges is set on moderation, but it can also be used for events and anything else that may be fitting.
+
+## What Community Collab isn’t
+
+- It isn’t an official affiliation with the Quilt project itself or any other community part of the program
```
gdude2002, line 17: The program _is_ an official Quilt project, but this reads as if that's not the case. You need to be more specific regarding what you're talking about in general in this document - whether the program itself or, as is the case here, the program's members.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+
+Managers are responsible for keeping the collab program up and running.
+
+- Maintain and organise the associated Discord server
+- Actioning changes to the collab-related spaces
+- Organise polls and votes as required by formal processes
+- Help resolve conflicts between communities
+- Stay in touch with any community part of the program
+
+Community Collab Managers are the same as Community Managers of the Quilt Community Team.
```
gdude2002, line 37: This statement doesn't make much sense in English. Are all Quilt CMs also Collab managers? Are Collab managers required to be Quilt CMs? Are the positions simply similar?

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Collaborators
+
+Collaborators are moderators of one or more communities in the program. They have access to the server to the moderation part of the server, such as general moderator chats and case threads. Access is also given to the event category.
```
gdude2002, line 41: This is word soup, and doesn't parse in my brain.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Event Planners
+
+Event Planners are members of the staff team of one or more communities in the program, who focus on event organisation. They have access to the events part of the server.
```
gdude2002, line 45: You stated "the event category" above, but used "the events part of the server" here. Consistency would be a good idea, I think - and "the events category" (with a plural "events") feels more formal and professional.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Project Roles
+
+Other roles may be given surrounding the different projects happening inside the program. Those roles are informal and will depend on each project.
```
gdude2002, line 49: Again, this really doesn't word well and could be confusing. It'd be better to go with something more formal, like...

> Other roles may be assigned based on the different projects that belong to the Collab program.

I also disagree with the notion that the roles are informal as they serve a purpose; it'd be better to explain why they don't need to be part of the RFC instead.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Invitees
+
+Invitees are temporary users, with no access to the rest of the server. They only have access to temporary channels required by the process they are visiting for.
```
gdude2002, line 53: This does not look like the consultant concept I put forward. Essentially, consultants would be collaborators that aren't representing specific communities; they'd have the same access, but would be there to take part in the Collab program without directly representing a group.

I don't see much value in this "Invitees" concept, and I don't believe it was discussed at all.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Veto rights
+
+While the vote is active, any Collaborator or Manager can use their veto right by sending a message inside the thread, detailing their reasoning.
+
+If another Collaborator or Manager believes the reasoning isn’t satisfying, they can start the overruling process. If so, another poll is set up, if more than half of the individual votes are positive after three days, the veto is discarded.
+
+If no Collaborator or Manager overrules the veto after three days, the vote is considered as failed right away. The Managers are free to only communicate the result to the Invitees at a later date to keep the veto secret.
+
+## Process: adding a community
```
gdude2002, line 81: Again, I feel like the terms you've used haven't been well defined in this block. For example, you use "community" a lot, but it's often unclear whether you mean the applicants (the community applying), or the program (the QCC members).

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+
+Any community, ideally having their main user space on Discord, centred around Minecraft, can apply to join the Community Collab program. The process is defined as follows:
+
+1. A member of the community holding administrative or moderation powers needs to get in touch with a Manager to start the process. The following questions should be asked to the community and collected, along with an invite to the server:
+    - What is the main focus of the server?
+    - Why is your community looking to join the Community Collab program?
+    - What is your current ruleset?
+    - Who is currently part of your staff team, and who would have access to Community Collab?
+2. Once those answers are collected, the Manager can inform all the Collaborators about the community, and start a discussion to gauge the interest of the community.
+3. The members of the community holding either administrative or moderation powers are invited to the Community Collab server under the Invitee role and a channel is set up to allow discussion between Collaborators and Invitee.
```
gdude2002, line 91: Why? Also, if the invitee idea only exists for this purpose, it should probably be folded into the process; though I don't see a reason to bring in more than one or two representatives from the community that's applying.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+4. Once the Collaborators informally agree that enough information has been gathered, the vote can be created in a private channel which cannot be seen by the Invitees, with the option to approve or deny, according to the process defined in this document. The Invitees should be informed the vote has begun.
```
gdude2002, line 92: You use the word "can" a lot in this document, but I think we should be much more assertive about these processes. "Can" always implies that it's not a requirement.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+- If the vote has passed: the Invitees are elevated to the rank of Collaborators. Internal discussions can be purged first if needs be. The community is added to any listing that may exist.
+- If the vote is rejected: enough time is given to the Invitees to acknowledge the result and leave the server. Managers are free to kick Invitees if necessary.
+
+# Process: removing a community
+
+The Managers can decide to start the process of removing a community from the program, due to external events or shared concern between communities. The process is defined as follows:
+1. The Managers can temporarily remove the community from the program by replacing roles of the affected members with the Invitee role, if they believe it is necessary to protect the other communities. The Managers are free to communicate the reason to the community, but it is not required.
+2. The Managers shall write a message clearly explaining the concerns, with an attached thread. In any case, the affected communities cannot have access to this thread. A channel can be set up to allow discussions and clarifications.
```
gdude2002, line 101: Aren't "the affected communities" everyone in QCC aside from the community in question?

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+- If the vote is rejected: the community is reintegrated to the program, and roles are given back if necessary. Internal discussions can be purged first if needs be.
+
+# Rationale and alternatives
+
+The Quilt Community Collab program existed prior to this document. This document tries to summarize how the server has been working so far while formalising some aspects that were still left up in the air.
+
+More importantly, three polls have been held inside the server, with the following results:
+
+- When voting for changes on who is or isn't part of the CC network, how should we count votes?
+    - 10 answered: One vote per community: each community will have to write down a formal process on which they decide to vote for or against changes. This process must be democratic and fair.
```
gdude2002, line 114: If we're requiring a documented, formal voting process for each community (which I don't think we actually need), it should be provided as part of a community's application. The RFC isn't clear on this, however.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+    - 4 answered: >50% approval to pass (there must be more approval than neutral)
+    - 1 answered: 100% approval to pass (there are no neutral votes, only vetos)
+- When voting for changes on who is or isn't part of the CC network, should we have veto rights, and if so how should they work?
+    - 17 answered: Any voter can veto, which can then be overruled by the majority if the concerns aren't believed to be deserved
+    - Nobody answered: There are no veto rights.
+    - Nobody answered: Any voter can veto, which will immediately cancel the vote
+
+# Prior art
+
+Quilt Community Collab is an indirect replacement of the Modding Interpol, a now extinct private server used to share threat info between communities, with a wider range of activities, such as events and related projects.
```
gdude2002, line 127: image

Y'all, we had a discussion about this, but I think you forgot to include it

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+
+# Unresolved questions
+
+The following questions will have to be solved at a later date:
+
+- We aren’t defining any role for administrators of the communities part of the program
+- We only allow for full communities to join, we should look at whenever we also want to invite individuals
```
gdude2002, line 134: See my comments about consultants

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+- We are locking Managers to be part of the Quilt Community Team, but we should look at whether we want to allow for other communities to have Managers
```
gdude2002, line 135: You addressed this in a resolved comment, but I don't think you updated the document

gdude2002 (changes requested):
I don't think you're going to like this review, haha

structure/0062-quilt-community-collab.md
Akarys42: Invitee is a temporary role used for specific process. For example, the members of a server who wish to join CC would get that role while the vote is active. This is different to the consultant role, which currently doesn't exist in this RFC, considering it hasn't ever been used. I'm not opposed to bringing it back though.

gdude2002: It was definitely my intention to get that nailed down in this RFC, as we do have potential consultants waiting.

structure/0062-quilt-community-collab.md
Akarys42: I don't see a reason to limit the amount of persons able to join as invitees. I'd rather meet all the people who are able to join the server later now, than discover them later. Also the invitee role can be used during the removal process, and may be used by future processes, such as individual collaborators.

structure/0062-quilt-community-collab.md
Akarys42: Right, that's a wording issue. Here it is used as communities who may get removed from the program.

structure/0062-quilt-community-collab.md
Akarys42: You need to give us more info than this lmao

structure/0062-quilt-community-collab.md
Akarys42: Consultants are a different idea as far as I can tell

gdude2002: No, that was the intention

structure/0062-quilt-community-collab.md
gdude2002: Huh? We had a discussion about it, don't y'all recall?

structure/0062-quilt-community-collab.md
Akarys42: I do believe that's something we may want to look at in the future. There would be some great advantages to not having CC 100% controlled by Quilt people. That being said, I'd rather wait a fair bit before looking at this.

structure/0062-quilt-community-collab.md
gdude2002: Hmm, okay, I see

structure/0062-quilt-community-collab.md
Akarys42: The way I understand it is Consultant is an additional role anyone already in the server can take. My proposal would be to allow individual users not part of any CC community to join the server after a vote.

structure/0062-quilt-community-collab.md
Akarys42: I don't, maybe someone will later on. That said, could you please just give a suggestion for this paragraph? I am definitely not knowledgeable enough here.

gdude2002: Basically, the name is wrong, and it specifically focused on moderation, moderation resources, and social/cultural context (such as how to approach things like plurality).

structure/0062-quilt-community-collab.md
gdude2002: My idea of a consultant is exactly what you described, and what I mentioned in my earlier comment about it - someone that's part of collab but not representing a community - and not temporary either.

Akarys42: That doesn't match at all the description in the welcome channel. Anyway, this needs further discussion, I think we should leave it there for now and come back to it once this RFC is merged and tested.

I'll clarify what the invitee role is for, it is more of a technical thing than anything else. Maybe it shouldn't even be in the RFC.

gdude2002: The welcome channel is irrelevant for new policy, haha

gdude2002:
I forgot to mention it in my review, apparently, so I'll mention it now.

This policy needs to make room for Quilt admins, as this is a Quilt space and the admins are responsible for providing oversight over the Community Team. If your goal is to nail down the entire program in policy, this can't be merged until that's done.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+- It isn’t a place to blindly synchronise punishments between communities
+    - Each community must decide on appropriate moderation action based on facts shown to them and their own rules and policies. Infringing this rule may get your community removed from the Community Collab program.
+
+## Roles inside the server
+
+### Keyholder
+
+The keyholder for this space is the same as for the Quilt Community Team. This role is described in the [RFC 0007 # Keyholders](https://github.com/QuiltMC/rfcs/blob/master/structure/0007-community-team.md#keyholders).
+
+### Managers
```
Akarys42, line 27: Note to self: managers should be given moderation power if it is really required.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,135 @@
...
+
+Polls must include a closing date of at least ten days. If required, the closing date can be pushed away by a maximum of a month, upon the informal agreement with the majority of the other communities. During that period, any Collaborator can use their veto right (see below), if allowed by the process.
+
+Once the poll is closed, the vote can be interpreted, depending on the specific process this vote is part of, if any.
+
+## Veto rights
+
+While the vote is active, any Collaborator or Manager can use their veto right by sending a message inside the thread, detailing their reasoning.
+
+If another Collaborator or Manager believes the reasoning isn’t satisfying, they can start the overruling process. If so, another poll is set up, if more than half of the individual votes are positive after three days, the veto is discarded.
```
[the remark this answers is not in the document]
Akarys42: It is extended until the three days vote is over

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,201 @@
...
+Discord INTERPOL had many similarities with the Quilt Community Collab Program, including many of its foci and members. However, it lacked the support it needed from larger communities and didn't have the backing of a mod-loader project, which made it difficult to grow to a size that would see it become more widely useful.
+
+Additionally, the Quilt Community Collab Program has a much broader focus --- attempting to educate its staff on running their communities, and providing spaces to talk about general community management, defensive moderation tooling, wider social issues, and event planning.
+
+# Unresolved questions
+
+For the sake of getting this process document completed within a reasonable amount of time, the following concerns will need to be addressed at a later date:
+
+* We currently haven't discussed the possibility of opening Program membership to non-moderational community staff
+* We haven't discussed how new Managers should be added, which is an issue due to the fact that all Program Managers are also Quilt Community Managers; a more representative level of power-sharing should be worked on
```
sciwhiz12, line 200: Assuming that some process exists whereby collaborators of the Program can become Program Managers, if a new Quilt Community Manager is elected by the process of the Quilt team, do they automatically assume a position as a Program Manager?

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,201 @@
...
+3. A Discord channel is set up to allow the administrators and moderators from the new community (who are to be invited and given the Invitee role) to talk with existing Program members regarding their application
+4. Once the voting-eligible Program members have informally agreed that enough information has been gathered, the vote may be created (with options to approve or deny the vote) --- it must be done in a channel the Invitees don't have access to, and follow the voting process outlined earlier in this document
+5. Once the vote has concluded, the discussion channel must be archived and kept read-only, so it can be reviewed later on if needed
+
+### Interpreting the vote
+
+**The vote has passed** when more than 75% of the votes cast at the time of closing are positive. Internal discussions can be removed as needed before the new community is given access to the rest of the server. The new community should also be added to any relevant listings.
```
sciwhiz12, line 126: Should there be a clause to additionally and explicitly require some % of the membership to have cast a vote? Or are we banking on the fact that within the mandated timeframe, communities will have cast a vote before closing?

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,201 @@
...
+
+At any point before the poll has closed, any Collaborator can use their right to veto, if allowed by the process. This right is described below.
+
+Once the poll has been closed, the vote should be interpreted as described by the process the poll is a part of.
+
+### The right to veto
+
+While a poll is open and accepting votes, any Collaborator or Manager can use their right to veto by sending a message to the attached thread, detailing why they'd like to veto the vote.
+
+If another Collaborator or Manager isn't satisfied with the reasoning behind the veto, they can attempt to overrule the veto. To start this process, a new poll is created to allow voters to decide whether to overrule the veto. This poll lasts for three days and, if more than half of the individual votes are positive, the veto is overruled. If not, the original vote should be treated as if it has failed.
```
sciwhiz12, line 90: This section does not adequately clarify what happens after a Collaborator or Manager exercises their veto. How much time do Collaborators or Managers have after such an exercise of veto power to make their objection and start the process to overrule the veto, before the original vote is declared in the negative due to the veto?

Does "voters" in relation to the veto overrule poll here mean the voting communities, whereby the members of that community decide to cast a single vote in accordance to the above section, or does it mean the Collaborators and Managers which each get a single, individual vote?

What happens to the original poll's closing date while a veto overrule poll is open? Is the original poll's remaining duration frozen until the results of the veto overrule poll is finished? If not, does it mean that if a veto overrule poll closes at a time past the original poll's closing date, then the original poll is also immediately closed?

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,201 @@
...
+
+The veto process may be started on behalf of another Manager or Collaborator. Additionally, the list of voters on that veto may be kept secret. This is intended to avoid exposing those that supported the veto (or that didn't) to negative reactions or bullying by other Program members.
```
sciwhiz12, line 92: When a veto process is started on behalf of another, are the original vetoers required to be named, or can they remain anonymous? If the latter, should there be a clause requiring their identities be transmitted in secret to Program Managers and held in confidence, for the purpose of recordkeeping?

Does "list of voters on that veto" pertain to the veto overrule poll?

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,201 @@
...
+## What the Program is not
+
+* An official, direct affiliation with the Quilt project, or any other community that participates in the Program
+* A space for building or coordinating the character assassinations of community members or prominent personalities
+* A space meant for blindly synchronizing punishments between communities
+  * Instead, each community must decide on appropriate moderation actions for their own spaces, which may be informed by the data they're given access to, but not blindly copied between communities --- Breaking this rule is likely to result in a community being removed from the Program
+
+## Program Roles
+
+As the Program exists largely on Discord, the roles of its members are directly represented by Discord server roles.
```
sciwhiz12, line 26: This looks to not resolve @gdude2002's comment about the inclusion of the Quilt Administrative Board as oversight on the Quilt Community Team (https://github.com/QuiltMC/rfcs/pull/65#issuecomment-1272516842), unless that is covered by the section on Project Roles or I'm otherwise mistaken.

(I'd argue that, since the Quilt Administrative Board does not have moderation duties in the community spaces, it would be inappropriate to give them the Collaborator role.)

sciwhiz12 (changes requested):

structure/0062-quilt-community-collab.md
Akarys42: This is currently undefined, let's not spend time on this right now.

structure/0062-quilt-community-collab.md
gdude2002: 1, I'll update the document to state that, if nobody attempts to overrule a veto, it will apply when the original vote closes

2. Whoever is able to vote on the original poll. As vetos happen at an individual (rather than community) level, I think this should be done using individual votes.

3. I'll clarify that a vote to overrule will prevent the original vote from finishing until the vote to overrule has finished.

structure/0062-quilt-community-collab.md
gdude2002: I think these are both pretty clear already via the explanation of intent? I guess I can add "started or"

structure/0062-quilt-community-collab.md
gdude2002: Quilt admins shouldn't have the collaborator role. I'm not sure what to do with them right now, though.

structure/0062-quilt-community-collab.md
gdude2002: I think ten days is more than enough time to ensure that communities have voted. It's likely safe to assume they don't want to vote on the poll if they haven't by this time.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,249 @@
...
+
+At any point before the poll has closed, any Collaborator can use their right to veto, if allowed by the process. This right is described below.
+
+Once the poll has been closed, the vote should be interpreted as described by the process the poll is a part of.
+
+**Note:** Individual Collaborators currently do not hold the right to vote.
+
+### The right to veto
+
+While a poll is open and accepting votes, any Collaborator or Manager can use their right to veto by sending a message to the associated thread, detailing why they'd like to veto the vote.
```
Forkk, line 109: Allowing individuals to veto seems like it would defeat the point of communities voting together. It effectively means any single person who disagrees with the rest of their community can stop the vote and move to voting on an individual basis.

Maybe we should require an entire community to agree to veto?

Forkk (changes requested):
Most of this looks good to me, but do I have some concerns about the veto process.

structure/0062-quilt-community-collab.md
Akarys42: I thought we agreed on adding community admins to the collaborator role?

gdude2002 (approved):
image

Scotsguy (approved):

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,250 @@
...
+
+Invitees are temporary users with no access to Program-related spaces, other than temporary areas that are created for whatever purpose they're visiting the respective space.
+
+## Voting process
+
+Votes may be called by Program Managers, where each of the communities participating in the Program represent the poll's voters. As the Program exists largely on Discord, a poll and an associated thread are created on the Program's Discord server, in a channel designated for this purpose.
+
+The poll should be as detailed as possible in order to help everyone to make an informed decision. Discussions related to the poll should happen in the associated thread, to try to keep everything together.
+
+Each community is assigned one single vote. Individual communities are responsible for coming up with their own process for determining their vote, but it must follow these rules:
```
Southpaw1496, line 91: I have spoken about my feelings on votes being per-community and not per-collaborator, and I know I am in the minority, but I will put them here for the record:
- Members of smaller communities get more votes. Quilt is twice the size of the next-largest community, so we essentially get less than half the say of anyone else, despite Community Collab being managed by us. Not that I think we should get **more** weight than others relative to our size, but we certainly shouldn't get less.
- People who are members of multiple communities get multiple votes. Among other things, that means more well-known or prestigious community members could get a louder voice than those who may only be part of a single community. It may be a little dramatic to say that this goes against Quilt's value of making sure everyone has an equal voice, no matter their status, but I think it's an important point to make.
- In addition to the size of the community determining the voting power of a collaborator, their voting system (which is not standardised) could also have an influence on how much say they have.
- We have a problem with whether individual collaborators get votes, well, there's an easy solution.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,250 @@
...
+* Why the applicant wishes to join the program
+* The applicant's personal views on social issues
+* The applicant's known approaches to moderation, and the communities they've worked with
+* How the applicant has behaved in the communities that the Program members have access to
+* How well-known the applicant is among Program members
+* Whether the applicant has any known, relevant history
+
+## Process: Removing a community
+
+While we hope that it won't ever be necessary, it's important that we have the ability to remove communities from the program when it becomes necessary.
```
Southpaw1496, line 180:
```suggestion
While we hope that it won't ever be necessary, it's important that we have the ability to remove communities from the program if need be.
```
"when" sounds inevitable, rather like Thanos. I changed the last "necessary" since it was repeated twice in the same sentance.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,250 @@
...
+
+**The vote has passed** when more than 50% of the votes cast at the time of closing are positive. The community is removed from the Program and any listings therein.
+
+Otherwise, **the vote has failed.** Internal discussions can be removed as needed. The community is reintegrated, and any roles that were removed are returned.
+
+## Process: Removing an Individual Collaborator
+
+While we hope that it won't ever be necessary, it's important that we have the ability to remove Individual Collaborators from the program when it becomes necessary.
+
+This process matches the process for removing a community.
```
Southpaw1496, line 211:
```suggestion
The process for removing an Individual Collaborator matches the process for removing a community.
```
I don't see the need for a heading and an introduction that's longer than the actual content of the heading, while also basically repeating what was just said.

Southpaw1496 (commented):
I diiiiiiiiid it.

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,250 @@
...
+
+Collaborators are given access to all moderation-related areas within Program spaces, such as general moderation chat channels and case-tracking threads. Additionally, all Collaborators are granted access to spaces dedicated to organizing events between participating communities.
+
+### Individual Collaborators
+
+The Individual Collaborator role is given to moderators that have joined the Program, but do not represent any specific community, which may be for any number of reasons. We believe that Individual Collaborators should be able to make use of the Program, and can contribute to it just how regular Collaborators can.
+
+Individuals are given the same level of access that regular Collaborators are given. However, Individual Collaborators are not given voting rights --- this is something that still needs to be discussed, but for the sake of simplicity and ease of implementation, this discussion has been deferred until later.
```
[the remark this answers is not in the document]
Akarys42: I am actually going to remove that sentence. Giving voting permissions to individual collaborators doesn't make sense, and the veto right is a good alternative.

structure/0062-quilt-community-collab.md
Akarys42: While I appreciate your concerns, we held a vote inside the collab server and the majority decided to go for one vote per community. This isn't something I would be comfortable changing now. We can always revisit it later with another poll if needs be.

image

structure/0062-quilt-community-collab.md
```diff
@@ -0,0 +1,246 @@
...
+
+The Collaborator role is given to moderators from the communities that are part of the Program, and who have joined a Program-related space in order to represent their communities. Administrators from these communities are welcome to join as Collaborators as well, if they wish.
+
+Collaborators are given access to all moderation-related areas within Program spaces, such as general moderation chat channels and case-tracking threads. Additionally, all Collaborators are granted access to spaces dedicated to organizing events between participating communities.
+
+### Individual Collaborators
+
+The Individual Collaborator role is given to moderators that have joined the Program, but do not represent any specific community. Individual Collaborators can make use of the Program and contribute to it just as regular Collaborators can.
+
+Individuals are given the same level of access that regular Collaborators are given. However, Individual Collaborators are not given veto rights but not voting rights.
```
wafflecoffee, line 69:
```suggestion
Individuals are given the same level of access that regular Collaborators are given, including veto rights but not voting rights.
```

gdude2002: voting rights are not "levels of access" or even "access" at all

gdude2002:
ICYMI, I had to force push this branch - some of the suggestions were needlessly nitpicky, made the document harder to read, or changed the meaning of a passage such that it was no longer correct. Amber had already merged them in last night, and I couldn't figure out how to revert multiple commits, so I apologize for the loss of tracked history.

gdude2002:
:tada:

PR #71, opened 2023-02-13
Remove "triage" wording from QSL rfc

TheGlitch76:
Because despite it being undefined, community managers seem to think it forces them to remove dev responsibilities of QSL team members.

### DIFF
--- a/structure/0068-qsl-teams.md
+++ b/structure/0068-qsl-teams.md
@@ -4,7 +4,7 @@ This RFC establishes and defines the structure, roles, and responsibilities of t
 
 <!-- For simplicity, a lot of the boilerplate not relevant to a team definition was left out of this RFC-->
 ## Explanation
-The QSL Team is made up of three subteams, QSL Core, QSL Triage, and Quilted Fabric API. One cannot only be a member of the "QSL Team", they must also be a member of at least one subteam.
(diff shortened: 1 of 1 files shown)


CheaterCodes:
We usually wouldn't put an FCP on small changes like this, but since I know that some poeple are still looking for better wording, I'll give them one FCP period to bring them up before we merge this.

EnnuiL (changes requested):
I object to this name change due to it causing an issue where there are now 2 QSL Teams: the root team and the subteam; This change has been already implemented on Discord due to the Triage naming being confusing, but in a rush to get it changed due to the threat of the Triage team being treated as literally triage, we now having a name that really needs to be changed if we want to talk about the post-subteam-merger organization;

A proposed name was "QSL Development Team" for the sub-team, which does sound interesting, but I do worry about the imposter's syndrome it could cause

EnnuiL:
As long as the conflict remains and no further discussion happens, we won't have a definitive conclusion

Akarys42:
Alright, I think the simplest way to about this is to leave three additional days for other ideas to come in, and vote on the different solutions. This will force this fairly minor problem to get resolved, and make this compliant with the rest of the RFCs.

PR #72, opened 2023-03-15
RFC 0072: Legal Team

Akarys42:
This adds an RFC for the upcoming Legal Team.

This PR assumes #47 is merged. 

Initial members: @Akarys42, @gdude2002 
Initial team lead: @Akarys42 

[Rendered view](https://github.com/Akarys42/rfcs/blob/patch-1/structure/0072-legal-team.md).

### DIFF
--- a/structure/0072-legal-team.md
+++ b/structure/0072-legal-team.md
@@ -1,0 +1,88 @@
+# RFC 4269: Legal Team
+
+This document has been made according to the process outlined in [RFC 0006](https://github.com/QuiltMC/rfcs/blob/main/structure/0006-governance.md#teams).
+
+# Summary
+
+The Legal Team is a standalone team in charge of handling legal matters to the best of its ability. This team will not require its members to be lawyers 
+or have any form of training, although this would be preferred. 
+
+# Motivation
+
+Historically, legal matters have been handled by various different people without any proper structure. The de-factor team who would handle particular 
+matters (eg. the Admin Board for registering Quilt as a non-profit) can feel unqualified to handle those issues and/or not have the bandwidth for it. 
+Moreover, someone handling legal matters leaving that de-facto team should not lead to them not handling those matters anymore, especially 
+considering the current scarcity of trained or knowledgeable members.
+
+Some laws may also require us to have a designated legal team, such as to designate a Data Protection Officer 
+([Art. 37 EU GDPR](https://gdpr-info.eu/art-37-gdpr/)).
+
+# Explanation
+
+## Responsibilities
+
+The team is responsible for the following:
+
+- Ensuring Quilt stays up to date with data regulation laws (eg. EU GDPR, COPPA)
+    - Keep the privacy policies, and terms of services up to date
+    - Designate a DPO
+    - Responding to data privacy requests
+    - Assisting the Infrastructure team in staying compliant
+- Reviewing contracts signed by the organisation
+- Handling registration of the organisation as a legal entity, if decided
+- Advising other teams on any legal issues that may arise
+- Answering questions and concerns from community members to the best of their ability
+- Writing guidelines on what changes to Quilt projects require legal verification and providing a process for this verification to happen
+- Write internal or public processes on how to handle requests
+- Any other legal matter not explicitly listed in this document
+
+## Legal Team Server
+
+To perform their activity, the team has a private server not viewable by any other team, for discussing problems needing the uttermost privacy. 
+
+While this deviates from the usual observable approach that makes sure every team can be monitored by at least one other independent team, 
+privacy is legally required to handle some matters, such as data privacy requests.
+
+## Requesting Changes
+
+In order to stay compliant with regulations and laws, the Legal Team has the power to block any Pull Request or community space change that
+could negatively affects compliance. This overrides any decision made by a Team Lead.
+
+If a conflict arises between the Legal Team and a Team Lead, the Admin Board can be summoned for arbitration. In arbitration, the 
+Admin Board's decision is final and overrides the Legal Team.
+
+The Legal Team also has the power to open high-priority issues on organisation repositories and request changes to be 
+made as fast as reasonably possible.
+
+## Communication
+
+To be able to communicate with users, Legal Team members are provided with an email address, connected to one or many distribution lists.
+
+## Transparency Reports
+
+Because of the lack of oversight, the Legal Team will be required to publish public reports twice as year, containing as many information on
+their activity as allowed by the applicable laws, and taking into account user privacy.
+
+# Drawbacks
+
+This team will most likely contain non-professional members. While their knowledge may still be extensive of some legal issues, we have to keep in 
+mind they aren’t lawyers.
+
+Due to legal needs, some if not all of the team members may have their real names appear publicly. Explicit approval from a team member will be 
+required before publishing.
+
+The ability to request changes may create some friction between developers and legal team members.
+
+# Rationale and Alternative
+
+Alternatively, other teams can sporadically handle legal matters. This has some issues, such as forcing people who aren’t motivated to handle those 
+matters, not centralizing the information and lacking the means required to stay compliant. The existence of this team is also legally required
+in some cases, as outlined above. 
+
+# Prior art
+
+This team is close in spirit to the idea of a legal department in a workplace, although without requiring legal training. It is worth noting that 
+the workplace definition of a legal department includes a large majority of employees who aren't lawyers, but some other qualifying background
+or field training.
+
+Transparency Reports are similar to [Discord's transparency reports](https://discord.com/tags/transparency-reports).
--- a/structure/0072-legal-team.md
+++ b/structure/0072-legal-team.md
@@ -28,7 +28,7 @@ The team is responsible for the following:
 
 - Ensuring Quilt stays up to date with data regulation laws (eg. EU GDPR, COPPA)
     - Keeping the privacy policies and terms of services up to date
-    - Designating a DPO
+    - Designating a Data Protection Officier (DPO)
     - Responding to data privacy requests
     - Assisting the Infrastructure team in staying compliant
 - Reviewing contracts signed by the organisation
--- a/structure/0072-legal-team.md
+++ b/structure/0072-legal-team.md
@@ -4,20 +4,20 @@ This document has been written according to the process outlined in [RFC 0006](
 
 # Summary
 
-The Legal Team is a standalone team in charge of handling legal matters to the best of its ability. This team will not require its members to be lawyers 
-or have any form of training, although this would be preferred. 
+The Legal Team is a standalone team in charge of handling legal matters to the best of its ability. This team will not require its members to be lawyers
+or have any form of training, although this would be preferred.
 
 It is made clear that this team will not be providing legal guidance but only advices to the best of their ability and team members cannot be held
 liable for their work in good faith as a Legal Team member.
 
 # Motivation
 
-Historically, legal matters have been handled by various different people without any proper structure. The de-facto team who would handle particular 
-matters (eg. the Admin Board for registering Quilt as a non-profit) may feel unqualified to handle those issues and/or may not have the bandwidth for it. 
-Moreover, someone handling legal matters leaving that de-facto team should not prevent them from handling those matters, especially 
+Historically, legal matters have been handled by various different people without any proper structure. The de-facto team who would handle particular
+matters (eg. the Admin Board for registering Quilt as a non-profit) may feel unqualified to handle those issues and/or may not have the bandwidth for it.
+Moreover, someone handling legal matters leaving that de-facto team should not prevent them from handling those matters, especially
 considering the current scarcity of trained or knowledgeable members.
 
-Some laws may also require us to have a designated legal team, such as to designate a Data Protection Officer 
+Some laws may also require us to have a designated legal team, such as to designate a Data Protection Officer
 ([Art. 37 EU GDPR](https://gdpr-info.eu/art-37-gdpr/)).
 
 # Explanation
@@ -44,10 +44,10 @@ The team is responsible for the following:
 In order to stay compliant with regulations and laws, the Legal Team has the power to block any Pull Request or community space change that
 could negatively affect compliance. This overrides any decision made by a Team Lead.
 
-If a conflict arises between the Legal Team and a Team Lead, the Admin Board can be summoned for arbitration. In arbitration, the 
+If a conflict arises between the Legal Team and a Team Lead, the Admin Board can be summoned for arbitration. In arbitration, the
 Admin Board's decision is final and overrides the Legal Team.
 
-The Legal Team also has the power to open high-priority issues on organisation repositories and request changes to be 
+The Legal Team also has the power to open high-priority issues on organisation repositories and request changes to be
 made as fast as reasonably possible.
 
 ## Communication
@@ -61,24 +61,24 @@ their activity as allowed by the applicable laws, and taking into account user p
 
 # Drawbacks
 
-This team will most likely contain non-professional members. While their knowledge may still be extensive of some legal issues, we have to keep in 
+This team will most likely contain non-professional members. While their knowledge may still be extensive of some legal issues, we have to keep in
 mind they aren’t lawyers.
 While legal liability may be a concern, the non-profit and its insurance will be here to provide some protection if an issue arises.
 
-Due to legal needs, some if not all of the team members may have their real names appear publicly, such as when signing bylaws. 
+Due to legal needs, some if not all of the team members may have their real names appear publicly, such as when signing bylaws.
 Explicit approval from a team member will be required before publishing a real name publicly and approval cannot be made mandatory.
 
 The ability to request changes may create some friction between developers and legal team members.
 
 # Rationale and Alternative
 
-Alternatively, other teams can sporadically handle legal matters. This has some issues, such as forcing people who aren’t motivated to handle those 
+Alternatively, other teams can sporadically handle legal matters. This has some issues, such as forcing people who aren’t motivated to handle those
 matters to do so, not centralizing the information and lacking the means required to stay compliant. The existence of this team is also legally required
-in some cases, as outlined above. 
+in some cases, as outlined above.
 
 # Prior art
 
-This team is close in spirit to the idea of a legal department in a workplace, although without requiring legal training. It is worth noting that 
+This team is close in spirit to the idea of a legal department in a workplace, although without requiring legal training. It is worth noting that
 the workplace definition of a legal department includes a large majority of employees who aren't lawyers, but some other qualifying background
 or field training.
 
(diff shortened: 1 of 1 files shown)


triphora (changes requested):
I would advise a name change. It's really dangerous to call a group a "legal team" if they have no legal experience and no rights to be providing legal counsel.

Akarys42:
We will do our best to get as much qualified people on the team as possible. As outlined in the RFC, we're still required by law to have this team, such as to provide a DPO. We cannot block the existence of that team just because we don't have a lawyer in Quilt, which isn't a reasonable expectation anyway. I'd note that we are still going to have some people with specific formal training (eg. on GDPR) and field experience with other subjects (eg. copyright laws or running a non-profit).

As for legal representation, it is something we'll only bother with and pay for if there is a need for it (eg. Quilt is being sued), which hopefully will never happen.

structure/0072-legal-team.md
```diff
@@ -0,0 +1,88 @@
...
+([Art. 37 EU GDPR](https://gdpr-info.eu/art-37-gdpr/)).
+
+# Explanation
+
+## Responsibilities
+
+The team is responsible for the following:
+
+- Ensuring Quilt stays up to date with data regulation laws (eg. EU GDPR, COPPA)
+    - Keep the privacy policies, and terms of services up to date
```
gdude2002, line 27: No comma needed here.

structure/0072-legal-team.md
```diff
@@ -0,0 +1,88 @@
...
+    - Designate a DPO
+    - Responding to data privacy requests
+    - Assisting the Infrastructure team in staying compliant
+- Reviewing contracts signed by the organisation
+- Handling registration of the organisation as a legal entity, if decided
+- Advising other teams on any legal issues that may arise
+- Answering questions and concerns from community members to the best of their ability
+- Writing guidelines on what changes to Quilt projects require legal verification and providing a process for this verification to happen
+- Write internal or public processes on how to handle requests
```
gdude2002, line 36: Bad tense here again

structure/0072-legal-team.md
```diff
@@ -0,0 +1,88 @@
...
+- Any other legal matter not explicitly listed in this document
+
+## Legal Team Server
+
+To perform their activity, the team has a private server not viewable by any other team, for discussing problems needing the uttermost privacy. 
+
+While this deviates from the usual observable approach that makes sure every team can be monitored by at least one other independent team, 
+privacy is legally required to handle some matters, such as data privacy requests.
```
gdude2002, line 44: We should discuss how we regulate this and assign users to it. Maybe this is also an infra team job? Potentially we need a separate infra team account for this? Or, perhaps, this should still be keyholder-relevant?

That discussion should be detailed here.

Alternatively, if that is not possible, we should discuss outsourcing this team to another organization somehow.

gdude2002 (changes requested):
There's definitely a lot of work to do here - some of my changes are nitpicky, but overall this feels like it's missing a lot of detail that would be useful here.

structure/0072-legal-team.md
```diff
@@ -0,0 +1,88 @@
...
+Some laws may also require us to have a designated legal team, such as to designate a Data Protection Officer 
+([Art. 37 EU GDPR](https://gdpr-info.eu/art-37-gdpr/)).
+
+# Explanation
+
+## Responsibilities
+
+The team is responsible for the following:
+
+- Ensuring Quilt stays up to date with data regulation laws (eg. EU GDPR, COPPA)
```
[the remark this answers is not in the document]
Akarys42: That's the official name of the regulation

structure/0072-legal-team.md
```diff
@@ -0,0 +1,88 @@
...
+While this deviates from the usual observable approach that makes sure every team can be monitored by at least one other independent team, 
+privacy is legally required to handle some matters, such as data privacy requests.
+
+## Requesting Changes
+
+In order to stay compliant with regulations and laws, the Legal Team has the power to block any Pull Request or community space change that
+could negatively affects compliance. This overrides any decision made by a Team Lead.
+
+If a conflict arises between the Legal Team and a Team Lead, the Admin Board can be summoned for arbitration. In arbitration, the 
+Admin Board's decision is final and overrides the Legal Team.
```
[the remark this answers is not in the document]
Akarys42: There is an odd number of admins to avoid that

structure/0072-legal-team.md
```diff
@@ -0,0 +1,88 @@
...
+This document has been made according to the process outlined in [RFC 0006](https://github.com/QuiltMC/rfcs/blob/main/structure/0006-governance.md#teams).
+
+# Summary
+
+The Legal Team is a standalone team in charge of handling legal matters to the best of its ability. This team will not require its members to be lawyers 
+or have any form of training, although this would be preferred. 
+
+# Motivation
+
+Historically, legal matters have been handled by various different people without any proper structure. The de-factor team who would handle particular
```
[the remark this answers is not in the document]
sciwhiz12: Could also be written as `*de facto*` ("The *de facto* team ..."), without the hyphenation and (optionally) italicized.

structure/0072-legal-team.md
Akarys42: We cannot legally justify having more people than strictly necessary seeing sensitive data. An overseeing body isn't strictly necessary according to the law. Usage of that server will be limited to only when legally required.

structure/0072-legal-team.md
```diff
@@ -0,0 +1,88 @@
...
+Because of the lack of oversight, the Legal Team will be required to publish public reports twice as year, containing as many information on
+their activity as allowed by the applicable laws, and taking into account user privacy.
+
+# Drawbacks
+
+This team will most likely contain non-professional members. While their knowledge may still be extensive of some legal issues, we have to keep in 
+mind they aren’t lawyers.
+
+Due to legal needs, some if not all of the team members may have their real names appear publicly. Explicit approval from a team member will be 
+required before publishing.
```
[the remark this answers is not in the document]
Akarys42: I'm not too sure what to say there. For example, if Quilt is registered as a non-profit, all team members would have to sign the bylaws, forcing their name to become public. There isn't much we can do about it other than make sure everyone is OK with that.

structure/0072-legal-team.md
```diff
@@ -0,0 +1,88 @@
...
+made as fast as reasonably possible.
+
+## Communication
+
+To be able to communicate with users, Legal Team members are provided with an email address, connected to one or many distribution lists.
+
+## Transparency Reports
+
+Because of the lack of oversight, the Legal Team will be required to publish public reports twice as year, containing as many information on
+their activity as allowed by the applicable laws, and taking into account user privacy.
```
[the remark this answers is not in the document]
Akarys42: Taking into account the amount of legal requests we received so far, a report over 3 months of activity will most likely have zeros all across the board. I guess if we always use the same document template that's not too big of an issue. I feel like this set of guidelines is beside the point of this RFC, though. The only real obligation is what is already written down, having some flexibility over the structure and the format would be nice, as long as we keep the same level of transparency.

structure/0072-legal-team.md
```diff
@@ -0,0 +1,88 @@
...
+- Handling registration of the organisation as a legal entity, if decided
+- Advising other teams on any legal issues that may arise
+- Answering questions and concerns from community members to the best of their ability
+- Writing guidelines on what changes to Quilt projects require legal verification and providing a process for this verification to happen
+- Write internal or public processes on how to handle requests
+- Any other legal matter not explicitly listed in this document
+
+## Legal Team Server
+
+To perform their activity, the team has a private server not viewable by any other team, for discussing problems needing the uttermost privacy.
```
TheGlitch76, line 41: You claim in a comment that this server is meant to be used only when neccesary, but this sentence states that it exists "to perform their activity" generally. I recommend something like this:

```suggestion
The Legal Team has a private server not viewable by any other team. This server may only be used for discussing problems which need the uttermost privacy. 
```

structure/0072-legal-team.md
```diff
@@ -0,0 +1,88 @@
...
+
+While this deviates from the usual observable approach that makes sure every team can be monitored by at least one other independent team, 
+privacy is legally required to handle some matters, such as data privacy requests.
+
+## Requesting Changes
+
+In order to stay compliant with regulations and laws, the Legal Team has the power to block any Pull Request or community space change that
+could negatively affects compliance. This overrides any decision made by a Team Lead.
+
+If a conflict arises between the Legal Team and a Team Lead, the Admin Board can be summoned for arbitration. In arbitration, the
```
TheGlitch76, line 51: What does the removal of staff process look like for the legal team? Can we call a vote to remove someone if they overstep their bounds?

TheGlitch76 (changes requested):

structure/0072-legal-team.md
Akarys42: That's defined in RFC 6

Akarys42:
Alright, I've seen a few people saying we need a lawyer. We don't have the funds for that, that's just how it be. I may have a solution in the future to have *some legal support*, but that's about it. Those are tasks we still need to carry out, not having a lawyer doesn't exempt us from those tasks. It is also worth noting that some people still have formal training on some points, I am for example certified by multiple organizations as a GDPR compliance expert, even if I'm not a lawyer.

triphora:
The simple solution: don't call it a legal team! "Oversight Committee" is a term I've heard used for these sorts of things.

Akarys42:
Overseeing isn't its only role. If someone has a better suggestion, we can rename it, but it really isn't that big of a deal.

CheaterCodes:
Do we need to merge the privacy and legal teams? Iunderstand that you would fit into both, but I don't really see why they need to be the same, correct me if I'm wrong.

In particular, I could understand why a privacy team would require their own server without oversight, but I'm not sure how relevant that would be for other legal matters (which could be overseen by Admins/CMs).

Akarys42:
> Do we need to merge the privacy and legal teams? Iunderstand that you would fit into both, but I don't really see why they need to be the same, correct me if I'm wrong.
> 
> In particular, I could understand why a privacy team would require their own server without oversight, but I'm not sure how relevant that would be for other legal matters (which could be overseen by Admins/CMs).

It is the similar tasks handled by the same people. There are other legal matters that might need privacy, such as a DMCA, or needing to exchange private information when signing bylaws.

The goal is to keep the usage of that server to the strict minimum. I don't think splitting the teams would even change anything there.

Akarys42:
If someone has a better name for the team that still encompasses all the tasks carried out, feel free to suggest it here.

I personally don't have any other idea, nor do I see any fundamental (legal or not) issue with the current name.

Akarys42:
Hello GitHub?

Anyway, y'all re-requested for review I guess.

structure/0072-legal-team.md
```diff
@@ -0,0 +1,96 @@
...
+## Legal Team Server
+
+The Legal Team has a private server not viewable by any other team. This server must only be used for discussing problems 
+which need the uttermost privacy, as required by law. This includes, for example, talking about an ongoing data request or DMCA takedown
+without any name being redacted.
+
+While this deviates from the usual observable approach that makes sure every team can be monitored by at least one other independent team, 
+privacy is legally required to handle some matters, such as data privacy requests.
+
+The server shall be owned by one of the Legal Team leads, who is in charge of managing the server.
```
sciwhiz12, line 51: Given that this precludes oversight of the Legal Team's decisions or actions by other teams/organization members, should there be a required (or at least highly recommended) minimum of two members in the Legal Team, such that there is some level of accountability with respect to each other?

structure/0072-legal-team.md
```diff
@@ -0,0 +1,96 @@
...
+## Transparency Reports
+
+Because of the lack of oversight, the Legal Team will be required to publish public reports twice as year, containing as many information on
+their activity as allowed by the applicable laws, and taking into account user privacy, similar to [Discord's transparency reports](https://discord.com/tags/transparency-reports).
+
+# Drawbacks
+
+This team will most likely contain non-professional members. While their knowledge may still be extensive of some legal issues, we have to keep in 
+mind they aren’t lawyers.
+While legal liability may be a concern, the non-profit and its insurance will be here to provide some protection if an issue arises.
```
sciwhiz12, line 77: > the non-profit and its insurance
Could you clarify this?

sciwhiz12 (changes requested):

structure/0072-legal-team.md
sciwhiz12: I'm thinking this from the angle of accountability and mutual oversight: having/recommending/requiring another person on the Legal Team means that:
 - If the first person is absent or otherwise preoccupied, there is another who can act as well;
 - Information is known to both (as much as applicable laws allow) rather than sequestered only to a single person;
 - Each member can [over]see the actions of the other, and give their recommendations/opinions whilst retaining privacy of the matter to within the Legal Team.

It might be too much to some, but I think there is some value to requiring mutual oversight within the team, since the Legal Team is otherwise unchecked in discussions/decisions by other teams (outside of the required transparency reports and Administrative Board arbitration).

Akarys42: I can add it as a recommendation if you really want, but that cannot be a strict rule. The legal team still has to function even if there is a single member in it.

sciwhiz12: Fair enough then.

OroArmor (dismissed):
Some meta questions that should be answered in this RFC:
 - [ ] How does the legal team fit into the current team structure? 
 - [ ] is it under community or is it a new "branch" that is separate? 
 - [ ] Who will be the members? 
 - [ ] Are all admins more or less a part of the team in some aspect as the heads of the quilt org?

Akarys42:
The Legal Team is a team like any other; that shouldn't need clarification.

The initial members shouldn't be in the RFC, according to #47.

The Administrative Board isn't part of the team de facto, I don't see why this would need to be mentioned in the RFC considering it isn't the case for any team.

OroArmor:
- This didnt really answer my question, but i think i understand
- Makes sense
- As the heads of the organization, the admin board may have to perform legal duties such as signing contracts for the org. If just one admin is on the legal team, but is then unavailable, what needs to happen, etc. This is an important thing to figure out before this is ready

Akarys42:
The whole point of having a legal team is to not rely on admins to perform legal duties

OroArmor:
> The whole point of having a legal team is to not rely on admins to perform legal duties

Because Admins are the owners of the project, they have legal responsibilites even if there is a legal team. You can't just hand wave this away. It needs to be address how the legal team and admin board interact

Akarys42:
The admins have no mandatory responsibilities as defined by the law. If admins want to ask to join that team they're more than welcome to, but making them a de-facto member makes no sense, especially considering the problems with having your legal name be known.

structure/0072-legal-team.md
```diff
@@ -0,0 +1,85 @@
+# RFC 0072: Legal Team
+
+This document has been written according to the process outlined in [RFC 0006](https://github.com/QuiltMC/rfcs/blob/main/structure/0006-governance.md#teams).
+
+# Summary
+
+The Legal Team is a standalone team in charge of handling legal matters to the best of its ability. This team will not require its members to be lawyers 
+or have any form of training, although this would be preferred.
```
sciwhiz12, line 8: On a point of formatting: some lines within this document have trailing spaces. I believe those should be removed.

structure/0072-legal-team.md
```diff
@@ -0,0 +1,85 @@
+# RFC 0072: Legal Team
```
sciwhiz12, line 1: On a point of formatting: all headers within this document other than this title header should be increased by one level, as there should only be one H1 header (this one).

sciwhiz12 (changes requested):

structure/0072-legal-team.md
Akarys42: Hard disagree - it makes the file much more navigable by only going from h1 to h3. It doesn't make sense to only have h2.

sciwhiz12: Can you please clarify for me what you mean? Are you saying that it is more navigable to have multiple H1 headers with a single nested level of H2 headers, compared to an H1 header for the title, H2 headers for each section, and a nested level of H3 headers (under the Explanation section)?

Based on a survey of all current RFCs, 9 out of 16 RFCs use H2 headers for the RFC sections, with 8 of those H2-using RFCs (and 1 of H1-using RFCs) using the H1 header as the title of the RFC.

If we expand the survey to all merged _and_ proposed RFCs, then 16 out of 26 use H2 headers, with 11 of these using H1 for the RFC title. The RFC template also uses H2 headers, with a single H1 header.

I therefore believe that, for better consistency with the majority of the current and proposed RFCs, the headings here should be modified so only one H1 header remains as the title, with sub-headings using H2 and lower-ranked headings.

LambdAurora (commented):
Was thinking about the name of this team and what about "Legal Advice Team"?
The bare name kinda bothers me as it makes it sound a bit too "professional"/"liable" (you get what you mean, I don't have the correct words to describe this). I feel like the "Advice" part would relieve that a bit.

Akarys42:
> Was thinking about the name of this team and what about "Legal Advice Team"? The bare name kinda bothers me as it makes it sound a bit too "professional"/"liable" (you get what you mean, I don't have the correct words to describe this). I feel like the "Advice" part would relieve that a bit.

For the most part, the team will be taking action and not just providing advice.

LambdAurora:
From my read of the RFC the work assigned to the team seemed to be more in the realm of advices, but sure, then I guess there's really no good name for this team.

Akarys42:
To clarify, the three main responsibilities of the team will be:
1. the NPO 
2. responding to DSR
3. keeping compliance

I can only see the last one being considered advice. Compliance Team or Clerical Team are the only two alternatives I could find, but I'm not a fan of those.

LambdAurora:
> I can only see the last one being considered advice. Compliance Team or Clerical Team are the only two alternatives I could find, but I'm not a fan of those.

Legal Compliance Team perhaps?

LambdAurora (commented):
I see you mention often about NPO but don't see the word anywhere in the RFC.
Did I missed something or I'm just not knowledgeable enough?

Akarys42:
> I see you mention often about NPO but don't see the word anywhere in the RFC. Did I missed something or I'm just not knowledgeable enough?

There it is

https://github.com/QuiltMC/rfcs/blob/463596947743fe0bd72c95c7c2261d221b6d8ab2/structure/0072-legal-team.md?plain=1#L35

LambdAurora:
Ok, was just me not being well-versed with acronyms then.

LambdAurora (approved):

sciwhiz12 (approved):

Akarys42:
This PR is placed under a 10 days FCP

Leo40Git (approved):

Cypher121:
I have to once again bring up the name. "Legal team" is an exceptionally loaded term for organisations. Having no legal counsel on it, in my view the arguments for using it anyway should be bulletproof and alternative names exhausted.

"Compliance team" as discussed above is better, I'd say. Yes, it doesn't cover registration as a non-profit and similar legal dealings. But I'd say those few situations would necessitate external legal advice to be done 100% properly to begin with.

Akarys42:
The majority of the Administrative Board voted in favor of the name Legal Team, this PR will therefore be merged as is.

PR #75, opened 2023-04-17
0062: Add a clause to call votes early

Akarys42:
If a vote doesn't allow vetos, once half of the communities voted the poll result is set in stone and cannot physically be changed. This avoids waiting for multiple days for absolutely no reason.


gdude2002:
Why was this merged? It hasn't even been here for 20 minutes

LambdAurora:
I have fucked up and accidentally merged instead of approving.
Good reminder to double check what you're clicking.

PR #76, opened 2023-04-17
RFC 0062: Add a clause to call votes early

Akarys42:
If a vote doesn't allow vetos, once half of the communities voted the poll result is set in stone and cannot physically be changed. This avoids waiting for multiple days for absolutely no reason.

### DIFF
--- a/structure/0062-quilt-community-collab.md
+++ b/structure/0062-quilt-community-collab.md
@@ -104,7 +104,7 @@ All polls must include a closing date that is set at least ten days into the fut
 
 At any point before the poll has closed, any Collaborator can use their right to veto, if allowed by the process. This right is described below.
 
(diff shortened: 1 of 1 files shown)


LambdAurora (approved):
I am so so sorry for having confused the approve button and the merge button.

OroArmor (approved):

gdude2002 (approved):

EnnuiL (approved):

0xf24 (approved):

LambdAurora:
This amendment being minor enough, as it cannot physically change the result of a vote and only shorten them when possible to allow for votes to relieve a bit of the weight of bureaucracy, it doesn't require a final comment period.

Enough reviews have been cast, so this amendment will be merged now.

PR #78, opened 2023-04-27
RFC 0007: Reduce vote length to 3 days

Akarys42:
Considering almost all votes are always casted within the first 72h and we are currently trying to cut down the length of our processes and the bureaucracy involved, I suggest to reduce the length of CT votes to 3 days.


gdude2002 (approved):

hqnna:
Feel like this doesn't allow for flexibility. Why not just make it so if the outcome has no chance of being overturned the vote is finalized?

Akarys42:
> Feel like this doesn't allow for flexibility. Why not just make it so if
the outcome has no chance of being overturned the vote is finalized?

Due to our thresholds being based on the number of casted votes, this is
quite hard to compute and I believe would require almost everyone to have
voted anyway.

Akarys42:
FCP until June 1st

PR #80, opened 2023-05-22
RFC 0001: Modernize the RFC process

Akarys42:
- Make the lifecycle a lead responsibility
- Shorten the FCP length to 3 to 7 days
- Allow FCPs to be used to force the discussion to end
- Allow minor FCPs to skip the process
- Clarify the Administrative Board as responsible for the RFCs without any team
- Remove dispositions
- Remove the need for universal approval

### DIFF
--- a/structure/0001-rfc-process.md
+++ b/structure/0001-rfc-process.md
@@ -10,7 +10,7 @@ This will allow us to put a time limit on unproductive discussion where necessar
 This document gives an outline of how the RFC process works and explains how contributors should submit their RFCs to the project. 
 
 ## When you must submit an RFC
-The RFC process requires that any large top-level organizational changes should be preceded by a "Request For Comments" document just like this one, explaining the proposed change, the motivation behind it, potential drawbacks, and other considerations. These documents are then discussed and revised by the community for some period of time until a maintainer decides that enough discussion has happened and begins a final comment period, after which a final decision is made.
+The RFC process requires that any large top-level organizational changes should be preceded by a "Request For Comments" document just like this one, explaining the proposed change, the motivation behind it, potential drawbacks, and other considerations. These documents are then discussed and revised by the community for some period of time until a leader of the responsible team decides that enough discussion has happened and begins a final comment period, after which a final decision is made.
 
 Most smaller changes to Quilt projects (such as bugfixes) can be submitted via the normal GitHub pull request system, however some are substantial enough that we would prefer that members of the community are given some time to discuss the potential change and give their feedback.
 
@@ -46,9 +46,9 @@ When you are ready to submit an RFC, the process will work as follows:
 - The team responsible for the RFC will discuss the RFC as much as possible in
   the pull request's comments. Discussion that occurs in other spaces should be
   summarized in the PR comments.
-- When the team responsible for the RFC decides that there has been enough discussion, a member of the team will issue a "motion for final comments". If the motion passes, it begins a 10 day period where the community has the last chance to give feedback on an RFC.
+- When the team responsible for the RFC decides that there has been enough discussion, a leader of the team will issue a "motion for final comments". If the motion passes, it begins a short period where the community has the last chance to give feedback on an RFC. This period must last at least 3 days and cannot last more than 7 days.
   + This step is taken when enough discussion has happened to make a final decision on the issue. It is designed to prevent RFCs from becoming stales.
-  + A consensus among all community members *is not necessary* for an RFC to
+  + A consensus among all community members *is not necessary nor expected* for an RFC to
     enter the final comment period. However, there should not be a strong
     consensus *against* the disposition given for the final comment period, and
     some argument supporting the decision should be clearly presented somewhere
@@ -56,12 +56,8 @@ When you are ready to submit an RFC, the process will work as follows:
   + If discussion has gone on for a long time, the team member who motions for
     final comments should post a summary comment outlining all of the
     discussion that has occurred.
-  + Before actually entering the final comment period, all members of the team
-    must sign their approval.
-  + When a team member motions for final comments, they also give a
-    "disposition" which can be either accept, reject, or postpone. This
-    determines what will happen to the RFC at the end of the final comment
-    period, although the disposition may change during that period.
+  + A FCP can be started to force the discussion to come to a conclusion.
+  + If the change is considered as "minor" by the team leads, the PR may skip the FCP process.
 - At the end of the final comment period, the RFC will be either accepted or
   rejected depending on its disposition at the close of the final comment
   period.
@@ -75,6 +71,8 @@ When you are ready to submit an RFC, the process will work as follows:
     + An RFC may also be "postponed". This means the RFC is effectively
       rejected at the moment because it does not fit into the project's
       near-term goals, but may be accepted at a later date.
+
+If there is no *de facto* team responsible for the RFC, the Administrative Board is considered as responsible.
 
 # Drawbacks
 - This RFC process will require constant moderation in order to ensure the
--- a/structure/0001-rfc-process.md
+++ b/structure/0001-rfc-process.md
@@ -50,10 +50,10 @@ When you are ready to submit an RFC, the process will work as follows:
   + This step is taken when enough discussion has happened to make a final decision on the issue. It is designed to prevent RFCs from becoming stales.
   + A consensus among all community members *is not necessary nor expected* for an RFC to
     enter the final comment period. However, there should not be a strong
-    consensus *against* the disposition given for the final comment period, and
+    consensus *against* the PR given for the final comment period, and
     some argument supporting the decision should be clearly presented somewhere
     in prior discussion.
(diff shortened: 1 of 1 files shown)


Earthcomputer (changes requested):
1. What is "a team leader"? Does this imply there can be multiple leaders on one team? Is there something in the RFCs to support this?
2. I don't like how this is continuing the trend to remove oversight from team members over their own projects. Unless this is in preparation for redefining team members to be something more broad that basically anyone can easily join, and making a "team leader" what team members used to be, I don't agree with this aspect of the change.

Akarys42:
Team leads are already properly defined as part of RFC 6. Not giving RFC power to individual developers make sense here.

structure/0001-rfc-process.md
```diff
@@ -10,7 +10,7 @@ This will allow us to put a time limit on unproductive discussion where necessar
 This document gives an outline of how the RFC process works and explains how contributors should submit their RFCs to the project. 
 
 ## When you must submit an RFC
-The RFC process requires that any large top-level organizational changes should be preceded by a "Request For Comments" document just like this one, explaining the proposed change, the motivation behind it, potential drawbacks, and other considerations. These documents are then discussed and revised by the community for some period of time until a maintainer decides that enough discussion has happened and begins a final comment period, after which a final decision is made.
+The RFC process requires that any large top-level organizational changes should be preceded by a "Request For Comments" document just like this one, explaining the proposed change, the motivation behind it, potential drawbacks, and other considerations. These documents are then discussed and revised by the community for some period of time until a leader of the responsible team decides that enough discussion has happened and begins a final comment period, after which a final decision is made.
```
Earthcomputer, line 13:
```suggestion
The RFC process requires that any large top-level organizational changes should be preceded by a "Request For Comments" document just like this one, explaining the proposed change, the motivation behind it, potential drawbacks, and other considerations. These documents are then discussed and revised by the community for some period of time until the leader of the responsible team decides that enough discussion has happened and begins a final comment period, after which a final decision is made.
```

structure/0001-rfc-process.md
```diff
@@ -46,22 +46,18 @@ When you are ready to submit an RFC, the process will work as follows:
 - The team responsible for the RFC will discuss the RFC as much as possible in
   the pull request's comments. Discussion that occurs in other spaces should be
   summarized in the PR comments.
-- When the team responsible for the RFC decides that there has been enough discussion, a member of the team will issue a "motion for final comments". If the motion passes, it begins a 10 day period where the community has the last chance to give feedback on an RFC.
+- When the team responsible for the RFC decides that there has been enough discussion, a leader of the team will issue a "motion for final comments". If the motion passes, it begins a short period where the community has the last chance to give feedback on an RFC. This period must last at least 3 days and cannot last more than 7 days.
```
Earthcomputer, line 49:
```suggestion
- When the team responsible for the RFC decides that there has been enough discussion, the leader of the team will issue a "motion for final comments". If the motion passes, it begins a short period where the community has the last chance to give feedback on an RFC. This period must last at least 3 days and cannot last more than 7 days.
```

Earthcomputer (changes requested):
I see, then the following changes are necessary. If only one member of each team is allowed to do this then you're going to have a bus factor problem.

Akarys42:
FCP starting today, ending on June 5th

structure/0001-rfc-process.md
```diff
@@ -46,22 +46,18 @@ When you are ready to submit an RFC, the process will work as follows:
...
   + This step is taken when enough discussion has happened to make a final decision on the issue. It is designed to prevent RFCs from becoming stales.
-  + A consensus among all community members *is not necessary* for an RFC to
+  + A consensus among all community members *is not necessary nor expected* for an RFC to
     enter the final comment period. However, there should not be a strong
     consensus *against* the disposition given for the final comment period, and
```
BucketOfCompasses, line 53: This still references disposition which have been removed as a requirement, should this be made more general e.g. there shouldn't be a strong consensus against the PR before an FCP can take place.

lukebemish:
Hmm, perhaps I'm missing it, but with the removal of the disposition thing how is it determined whether or not ab RFC is accepted? I'm a bit confused on some of the details of that process. Is it just a vote by the relevant team?

BucketOfCompasses:
> Hmm, perhaps I'm missing it, but with the removal of the disposition thing how is it determined whether or not ab RFC is accepted? I'm a bit confused on some of the details of that process. Is it just a vote by the relevant team?

I believe the removal of disposition means that the team will not need to tell you what action they intend on taking when the FCP starts, e.g. if they want to reject it, defer, or want to accept it, however when the period end they will give you their verdict then.

structure/0001-rfc-process.md
```diff
@@ -46,22 +46,18 @@ When you are ready to submit an RFC, the process will work as follows:
...
     some argument supporting the decision should be clearly presented somewhere
     in prior discussion.
   + If discussion has gone on for a long time, the team member who motions for
```
BucketOfCompasses, line 56: team member here should probably be changed to team leader

Oliver-makes-code:
> Allow FCPs to be used to force the discussion to end
Wouldn't that defeat the purpose of a final comment period?

lukebemish:
> I believe the removal of disposition means that the team leader will not need to tell you what action they intend on taking when the FCP starts, e.g. if they want to reject it, defer, or want to accept it, however when the period end they will give you their verdict then.

Hmm. I'm a bit confused about some of the changes here then, I think - knowing whether or not your comments are relevant during an FCP would seem to require knowing what the likely result of that FCP is, right? And sometimes that's very hard to tell if there's not much discussion yet. Though I may have missed the point here. I'm also a bit worried about the whole "using FCP to force discussions to end" thing. I think I get what it's trying to do, but some more clarification should probably be added, especially with the whole bit about FCPs no longer having a disposition attached - is the discussion being forced to end because a likely decision won't be changed by further discussion, or because there's enough discussion that it won't be accepted at present? Maybe this isn't necessary, but I've seen a few RFCs lately with comments just saying basically "This RFC is now in FCP" which I'm never quite sure how to read.

Akarys42:
Basically, this aims at changing the way FCP works to better match what we're already doing right now.

Instead of saying "this is what we'll do" and having a 10 days buffer for anyone to disagree, we can instead say "this PR will be merged in 5 days or more, so get your feedback in right now". This reduces a lot of the burden of FCP processes, and allow PRs without any feedback to continue moving forward.

lukebemish:
Hmm. Okay, so, if I understand this right: after an FCP period the PR will always be merged? That makes a lot of sense and does match what's going on now so I think I understand it better now. My one question is: how are PRs rejected under this system?

Akarys42:
Nah, after the FCP the PR is either merged, closed, or delayed. It is up to the team lead to do any of those three based on the discussions in the PR before and during the FCP.

lukebemish:
Ah, okay. I *think* I understand that change now

lenrik1589 (commented):
Also i believe that RFC amendments, such as this one, should be their own RFCs, not unlike IETF RFCs for example [RFC7405](https://www.rfc-editor.org/rfc/rfc7405) amending [RFC5234](https://www.rfc-editor.org/rfc/rfc5234) by including  
"Obsoletes: [5234](https://www.rfc-editor.org/rfc/rfc5234)" in RFC7405   
and  
"Updated by: [7405](https://www.rfc-editor.org/rfc/rfc7405)" in RFC5234

Akarys42:
> Also i believe that RFC amendments, such as this one, should be their own RFCs, not unlike IETF RFCs for example [RFC7405](https://www.rfc-editor.org/rfc/rfc7405) amending [RFC5234](https://www.rfc-editor.org/rfc/rfc5234) by including "Obsoletes: [5234](https://www.rfc-editor.org/rfc/rfc5234)" in RFC7405 and "Updated by: [7405](https://www.rfc-editor.org/rfc/rfc7405)" in RFC5234

We use the Git history for that, in order to keep RFCs somewhat readable

structure/0001-rfc-process.md
```diff
@@ -46,22 +46,18 @@ When you are ready to submit an RFC, the process will work as follows:
...
     final comments should post a summary comment outlining all of the
     discussion that has occurred.
-  + Before actually entering the final comment period, all members of the team
-    must sign their approval.
-  + When a team member motions for final comments, they also give a
-    "disposition" which can be either accept, reject, or postpone. This
-    determines what will happen to the RFC at the end of the final comment
-    period, although the disposition may change during that period.
+  + A FCP can be started to force the discussion to come to a conclusion.
```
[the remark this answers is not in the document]
Akarys42: Some RFCs do not get any activity or enter a stale state without any further progress. In this case, an FCP can be used to force the discussion to continue and come to a conclusion, even if it ends up being pushing the PR to a later date. Requiring full approval is impossible, we have too large teams. I do not believe abuse is possible, as RFCs still require a public announcement and at least 3 days of waiting for any potential opposition.

OroArmor (approved):

PR #81, opened 2023-06-01
RFC 0081: Active User Beacon 

Akarys42:
This RFC describes a privacy-friendly mechanism through which the loader can signal themselves as an active user, in order to provide a MAU (Monthly Active User) metrics. 

If implemented, an accompanying blog post will be published before release, clearly explaining the process and opt-out methods.

This RFC as been validated by the Legal Team as not requiring a privacy policy change. 

[Preview](https://github.com/QuiltMC/rfcs/blob/active-user-beacon/specification/0081-active-user-beacon.md)

### DIFF
--- a/specification/0081-active-user-beacon.md
+++ b/specification/0081-active-user-beacon.md
@@ -1,6 +1,6 @@
 ## Summary
 
-This RFC describes a privacy-friendly mechanism through which the launcher can signal themselves as an active user, in order to provide MAU (Monthly Active User) metrics. 
+This RFC describes a privacy-friendly mechanism through which the loader can signal themselves as an active user, in order to provide MAU (Monthly Active User) metrics. 
 
 ## Motivation
 
--- a/specification/0081-active-user-beacon.md
+++ b/specification/0081-active-user-beacon.md
@@ -10,7 +10,7 @@ The choice to not include any metadata in the signal was made to keep the mechan
 ## Explanations
 
 On launch, the loader will start the following process asynchronously:
-1. Check if the `loader.disable_beacon` property or the `LOADER_DISABLE_BEACON` environment variable is set to `true`. If so, the process is aborted (the launch process is unaffected).
+1. Check if the `loader.disable_beacon` property or the `QUILT_LOADER_DISABLE_BEACON` environment variable is set to `true`. If so, the process is aborted (the launch process is unaffected).
 2. Check for the last signalled month in a standard persistent location, if it is equal to the current month the process is aborted.
 3. Send a `POST` request to `https://beacon.quiltmc.org/signal` without any specific body or headers.
 4. Save the current month into the same standard persistent location as step 2.
--- a/specification/0081-active-user-beacon.md
+++ b/specification/0081-active-user-beacon.md
@@ -40,4 +40,4 @@ It is also worth noting all of these provide download statistics and not MAU cou
 
 ## Privacy Statement
 
-Nor in this RFC or in any further RFC is Quilt going to implement data collection without explicit user consent and without any functionality loss if said consent isn't given.
+Nor in this RFC or in any further RFC is Quilt going to implement data collection without explicit user consent and without any functionality loss if said consent isn't given, unless said functionality technically requires the data to be collected.
--- a/specification/0081-active-user-beacon.md
+++ b/specification/0081-active-user-beacon.md
@@ -10,7 +10,7 @@ The choice to not include any metadata in the signal was made to keep the mechan
 ## Explanations
 
 On launch, the loader will start the following process asynchronously:
-1. Check if the `loader.disable_beacon` property or the `QUILT_LOADER_DISABLE_BEACON` environment variable is set to `true`. If so, the process is aborted (the launch process is unaffected).
+1. Check if the `loader.disable_beacon` property or either of the `QUILT_LOADER_DISABLE_BEACON` and `CI` environment variables is set to `true`. If so, the process is aborted (the launch process is unaffected).
 2. Check for the last signalled month in a standard persistent location, if it is equal to the current month the process is aborted.
 3. Send a `POST` request to `https://beacon.quiltmc.org/signal` without any specific body or headers.
 4. Save the current month into the same standard persistent location as step 2.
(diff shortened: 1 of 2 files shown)


specification/0081-active-user-beacon.md
```diff
@@ -0,0 +1,43 @@
+## Summary
+
+This RFC describes a privacy-friendly mechanism through which the launcher can signal themselves as an active user, in order to provide a MAU (Monthly Active User) metrics.
```
ix0rai, line 3:
```suggestion
This RFC describes a privacy-friendly mechanism through which the launcher can signal themselves as an active user, in order to provide MAU (Monthly Active User) metrics. 
```
grammar fix

ix0rai (approved):

specification/0081-active-user-beacon.md
```diff
@@ -0,0 +1,43 @@
...
+
+When discussing with partners, requesting grants or negotiating sponsorships, it is quite important for The Quilt Project to be able to give accurate MAU metrics. 
+The choice to not include any metadata in the signal was made to keep the mechanism privacy-friendly, as Quilt collects zero personal data or usage statistics.
+
+## Explanations
+
+On launch, the loader will start the following process asynchronously:
+1. Check if the `disableBeacon` property or the `QUILT_DISABLE_BEACON` environment variable is set. If so, the process is aborted (the launch process is unaffected).
+2. Check for the last signalled month in a standard persistent location, if it is equal to the current month the process is aborted.
+3. Send a POST request to `https://beacon.quiltmc.org/signal` without any specific body or headers.
```
OroArmor, line 15:
```suggestion
3. Send a `POST` request to `https://beacon.quiltmc.org/signal` without any specific body or headers.
```

OroArmor (changes requested):

TheEpicBlock:
Using the term "launcher" here is a bit confusing, I was wondering what MultiMC/Prism/etc had to do with this. Later in the document it mentions loader, maybe just use that everywhere?

Akarys42:
> Using the term "launcher" here is a bit confusing, I was wondering what MultiMC/Prism/etc had to do with this. Later in the document it mentions loader, maybe just use that everywhere?

It was a typo, fixed in 04f42cf

specification/0081-active-user-beacon.md
```diff
@@ -0,0 +1,43 @@
...
+
+The server hosted at `beacon.quiltmc.org` MUST:
+1. Be open-source.
+2. Be auditable by any Quilt developer who wish to, within reason.
+3. Not store any data present inside the request.
+4. Not process any data present inside the request beside what is strictly required to establish the connection, following current best practices and technical feasibility. 
+
+## Drawbacks
+
+This process can be seen as privacy-breaking by some. We implemented an opt-out and decided to not collect any data to minimize some of that concern.
```
ImperatorStorm, line 30: Shouldn't this be opt-_in_?

hibiii:
This sounds super reasonable but I do raise a minor concern: would there be any deduplication mechanism on the server side?

I ask because I have multiple Quilt instances on my launcher and they do not share folders, I can foresee my instances pinging the Quilt servers more than once per month. I don't want to artificially inflate the numbers, but I presume this is a common occurrence across tinkerers and developers as well as people who play separate mod packs, given the popularity of launchers that separate instance folders by default, such as MultiMC and derivatives (like Prism).

specification/0081-active-user-beacon.md
hibiii: If it was opt-in then you wouldn't count the active users. You'd count the people that care enough about Quilt to report their own presence, and that's a much smaller group, you'd vastly underreport the number of active users.

AlexIIL:
@hibiii Since the RFC mentions a `...a standard persistent location` I assume that means something like `<user>/.quilt_loader/beacon.txt`, rather than in the instance folder itself. As such we wouldn't need to deduplicate requests on the server.

Akarys42:
Yes, the goal is to have the persistent file be user scoped when possible

OroArmor (approved):

AlexIIL (approved):

Scotsguy:
If we're already storing the last sent month in a file, would it be possible to add that as an opt-out method as well? I.e. if the content of the file is not a number, but "opt out"¹ or similar, just abort like if the env variable was set.

¹ Though aborting should be what happens on any error condition anyway (such as failing to read the month), I think it should still be explicitly mentioned

Akarys42:
We will not implement that, to reduce the likelihood of an external tool deciding to opt-out for the user without their approval. An invalid date inside the file will be ignored by the loader and the signal will be sent as normal, and the date fixed.

johannes-schmatz:
> something like `<user>/.quilt_loader/beacon.txt`

about this implementation detail:
It would be nice if it would respect the [XDG specification](https://en.wikipedia.org/wiki/Freedesktop.org#User_directories), so that it ends up in something like `~/.cache/quilt_loader/beacon.txt`, as that would keep the home directory free of any clutter. Idk if this should be put in the rfc directly or not.

AlexIIL:
> It would be nice if it would respect the [XDG specification](https://en.wikipedia.org/wiki/Freedesktop.org#User_directories), so that it ends up in something like ~/.cache/quilt_loader/beacon.txt, as that would keep the home directory free of any clutter. Idk if this should be put in the rfc directly or not.

Good point! Although I *think* this would end up in `XDG_CONFIG_HOME` *(`~/.config`)* instead.

Akarys42:
I appreciate the concern, although that's an implementation detail, it doesn't need to be on the RFC.

LambdAurora (approved):

johannes-schmatz (approved):

Fnige:
> We will not implement that, to reduce the likelihood of an external tool deciding to opt-out for the user without their approval. 

There should probably be a way to opt out still

LambdAurora:
> > We will not implement that, to reduce the likelihood of an external tool deciding to opt-out for the user without their approval.
> 
> There should probably be a way to opt out still

The PR describes ways to opt out, just not in a way that an external program can make the choice for the user.

hibiii (approved):

AlexIIL:
Final Comment Period ends on June 14th.

kb-1000:
How would this work with per-instance sandboxing? Loader presumably wouldn't just get write access (if any access at all) to a global ~/.config then.

Akarys42:
Instance sandboxing will most likely have to give the game access to a shared global folder. We can adjust once that's defined.

PR #82, opened 2023-06-20
Kill community centric RFCs

Akarys42:
While Quilt was initially trying to preserve its structure through RFCs, we've discovered this isn't an effective way to achieve this goal. Those RFCs have historically always been inaccurate compared to what's in action, because of the inability to move and iterate quickly due to all the legal-ish wording, procedures, and RFCs.

We've decided to give ourselves back the ability to be more agile with our structure, without loosing the core concepts that make up Quilt.


LambdAurora (approved):

mazziechai (approved):

gdude2002 (approved):

LemmaEOF (approved):

0xf24 (approved):

Akarys42:
This PR is placed in a FCP lasting 3 days, until June 24th.

TheGlitch76:
Renaming the files will kill a lot of links, like this one on [quiltmc.org](https://quiltmc.org/en/staff/moderation/moderation/).

Shim files should probably be left in their original locations explaining that the RFC in question has been archived and is no longer in effect.

Akarys42:
Do we really care? I don't think we do. Leaving the file in the original location makes the tree view quite inaccurate.

pluiedev (approved):

theRookieCoder:
I noticed that the community team and legal team RFCs have been archived, so what would this PR mean for the creation of new development teams? (I'm interested in the installer team in this case)

Akarys42:
This doesn't affect development teams

Oliver-makes-code (approved):
Looks fine

PR #83, opened 2023-07-04
Formalize Quilt mod metadata handling

EnnuiL:
With plans to introduce support for `quilt.mod.json5` (through [quilt-loader#329](https://github.com/QuiltMC/quilt-loader/pull/329) and [quilt-loom#36](https://github.com/QuiltMC/quilt-loom/pull/36)) in a limited capacity within the toolchain, formalizing how metadata should be handled has been needed. This formalizes the mandatory need to handle `quilt.mod.json` and the optional need of handling `quilt.mod.json5`

Considering that we are on a "down with bureaucracy" arc, I'd rather have this RFC be an explainer of things already been put in place, rather than as a precondition before we can move forward with QMJ5. I will, however, consider it as a precondition for further pushing of the format, with Quilt Template Mod waiting for a proper unveiling of the format before adopting it itself

### DIFF
--- a/rfc/0000-quilt-mod-metadata-handling.md
+++ b/rfc/0000-quilt-mod-metadata-handling.md
@@ -1,0 +1,33 @@
+# Quilt mod metadata handling
+
+## Summary
+
+This RFC specifies how the metadata of Quilt mods should be handled by its own toolchain, with the focus being on loader, build system and third-parties in the ecosystem.
+
+
+## Motivation
+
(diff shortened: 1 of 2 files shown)


EnnuiL:
Closing this, considering that this piece of bureaucracy, despite how surprisingly well-written it is, is no longer necessary;

It can be reconsidered later as a descriptive specification if needed, but in the hindsight? This RFC was only required to be written because someone had maliciously used bureaucracy as a personal veto to the whole project.

PR #87, opened 2023-11-07
Remove legal team

Southpaw1496:
(This pull request represents my personal views, which do not necessarily reflect the views of Quilt or other staff members)

## Motivation

The legal team was created at a time when Quilt was attempting to become more structured and "official" by doing things like registering as a nonprofit organisation. These efforts greatly increased the amount of bureaucracy that we had to deal with, and so a dedicated team made sense.

It has since become clear that there are significant drawbacks to becoming a more traditional organisation, in the form of the aforementioned bureaucracy, and greatly increased complexity in general. As well as this, the theoretical benefits of being a legal entity, such as grants and a lack of liability in some cases, aren't things that we actually need in practice. What we need is a reduction of complexity, and I think that removing the legal team is a step in that direction.

## Compliance and Data Protection without the legal team
The current RFC states that "some laws may also require us to have a designated legal team, such as to designate a Data Protection Officer [for the GDPR]". However, Article 37(1) states that a Data Protection Officer (or DPO) is only required in cases where:

> (a) the processing is carried out by a public authority or body, except for courts acting in their judicial capacity;
> 
> (b) the core activities of the controller or the processor consist of processing operations which, by virtue of their nature, their scope and/or their purposes, require regular and systematic monitoring of data subjects on a large scale; or
> 
> (c) the core activities of the controller or the processor consist of processing on a large scale of special categories of data pursuant to Article 9 or personal data relating to criminal convictions and offences referred to in Article 10.

Quilt is not a public authority, we do not systemically monitor data subjects on a large scale, and we do not process data constituted as "Special Data" by Article 9. As such, we do not need a Data Protection Officer.

## Delegation of the Legal Team's responsibilities
I believe that all the Legal Team's responsibilities will no longer need doing frequently enough to warrant a dedicated team. Many of the tasks they were responsible can be performed by existing teams when necessary:

- **Ensuring Quilt stays up to date with data regulation laws (eg. EU GDPR, COPPA)**:
    - **Keeping the privacy policies and terms of services up to date**: This can be performed by the Infrastructure team, as changes in privacy policy will usually only need to occur when the infrastructure changes.
    - **Designating a Data Protection Officier (DPO)**: We don't need one of these (see above).
    - **Responding to data privacy requests:** This can be performed by the infrastructure team, since accessing and deleting data requires access to infrastructure.
    - **Assisting the Infrastructure team in staying compliant:** The infrastructure team can self-regulate: compliance at Quilt's scale isn't difficult. The team will also have oversight from the Admin Board, as always.
- **Reviewing contracts signed by the organisation:** A dedicated team is not required for this. We can seek external advice if necessary.
- **Handling registration of the organisation as a legal entity, if decided:** We have no plans for this in the near future.
- **Advising other teams on any legal issues that may arise:** Teams can do their own research, or we can seek external advice if necessary.
- **Answering questions and concerns from community members to the best of their ability:** This can be handled by the appropriate teams on a case-by-case basis.
- **Writing internal or public processes on how to handle requests:** This only has to be done once, and does not require a dedicated team.

It's important to consider that, although members of the legal team have a knowledge of the law, they don't have (and have never had) proper legal qualifications, and so can never give "legal advice" in the proper sense. The RFC itself states:

> It is made clear that this team will not be providing legal guidance but only advices [sic] to the best of their ability

As such, when their responsibilities involved giving advice, reviewing contracts etc., not having a dedicated legal team to give this advice doesn't deprive us of proper legal advice, as no-one on the legal team was qualified to give that kind of advice in the first place. It is conceivable that we could have a legal team formed of people who *are* qualified to give legal advice, however, I think the need for it is infrequent enough that it would be better to simply seek advice externally if we ever need it.


Pyrofab:
Agreed with all your points, and I would add that empirically, relying on volunteers with no legal background to formally handle legal matters seemed to have a negative impact as their title inherently gives some credibility to their advice (which may be more easily questioned when coming from e.g. an infra team member).

With that said, maybe we should write down  explicitly somewhere that the infra team has the responsibility to keep up with data regulation laws ?

Southpaw1496:
I'm working on an update to all the privacy policies, I can definitely add that.

I'm just waiting for some clarifications from CHS (our host) before I publish them.

simpledemonn:
Law student here, interested in this team.

I am writing this following my messages in Discord, with the help of a friend. I believe removing this team would be a mistake.

Currently, Quilt is behind on multiple law related tasks, such as updating the privacy policy or GDPR compliance. Something that put the organization, and the individuals behind it, under some important risks. While, for example, the PR to update the privacy policies is open, it is my understanding Quilt does not currently have enough people on deck to actually see those mandatory tasks through, in the meantime leaving some very problematic holes such as claiming the data is stored in Europe, while from my research it appears to be stored in the US.

I am of the opinion this team needs a restructuration and to separate itself from its past, including renaming it if needed. While some tasks previously assigned to the team would not be relevant anymore, such as the non-profit as Quilt doesn't appear to be in a state where it would benefit from a proper legal structure (although this option should be kept on the table, should the project recover), keeping the legal side of things (such as privacy policies) checked, advising the infrastructure team on multiple matter and helping fill legal requests (such as the ones related to the GDPR), still exist.

An important point was made that this could be handled by existing teams. I disagree, as you are forcing anyone helping with said legal matters to have already joined the organization through another mean, while people like me who would be open to contributing, don't necessarily want to also have to contribute differently just to be able to work on what seems interesting to them.

Another point was made that an organization such as Fabric do not have a legal team, to which I would point to the fact they do not appear to be GDPR-compliant, as I could not find any privacy policy for the data collected through their services, and their claims about the protection provided by mappings are extremely dubious to me, which I believe several people at Quilt already agree to.

One last point I would like to make is, while it wouldn't be economically viable to pay a lawyer to work on the project, other means can easily be explored, such as getting help from students like me who are interested in the experience. This team wouldn't be Quilt's lawyers and therefore should not be legally responsible for their advises, and I can hear those arguments toward renaming the team, not deleting it.

tl;dr: I am of the opinion this team should stay to allow knowledgeable individuals to help with legal related discussions

simpledemonn:
> You are correct that the current policies are out-of-date, and I'm aware of the issues with them.

There are still multiple issues with the current privacy policies. I am not sure if these issues are identified or not yet, but it appears to me Quilt lacks the manpower to handle those right now.

> I never referred to Fabric specifically

I was referring to points raised in the Discord, sorry if that wasn't clear.

> you seem to be implying that projects like ours need a legal team to be GDPR-compliant, which is not true. We're not GDPR-compliant because our privacy policies are out-of-date (and I am working to get that fixed), but that doesn't mean we need a legal team.

Correct. What I am saying is it seems like you lack either the manpower and/or knowledge to do so, but feel free to correct me if I'm wrong.

> Even if a legal team was needed, I think we'd be very careful about accepting people into it who didn't have a pre-existing relationship with Quilt, especially given the powers that the team has, and the fact that they can't take responsibility for the advice that they give.

I haven't noticed the special powers this team has, I would be for removing them, so it only becomes an advisory team. Rather, a way to get input and things done from inside the organization. I don't see how this can be worse than the alternative you are suggesting, others contributing through pull requests or other means while working on partial information. 

> I don't deny that many of the duties that the legal team has aren't relevant (in fact I specifically addressed it), I just think that they can be done by other teams.

Again, is it seems like you lack either the manpower and/or knowledge to do so, but feel free to correct me if I'm wrong.

lukebemish:
Correct me if I'm wrong, but if the legal team:
- does not have any power within the quilt project, and is solely advisory
- has no legal responsibilities for advice given or actions taken
- does not require having some broader association with the quilt project
Why would such a team need to exist? Isn't that equivalent to just soliciting community feedback about privacy/whatever related issues? It could be I've misunderstood you.

aaronliu0130:
All the privacy policies haven't been updated since May. In fact, the legal stuff that were removed after the paperwork error are still up on the "Legal" page of the Quilt website. Are we sure we have enough legal resources?

PR #89, opened 2024-01-09
Support Sided Mixins

OroArmor:
See https://github.com/QuiltMC/quilt-loader/pull/395

### DIFF
--- a/specification/0002-quilt.mod.json.md
+++ b/specification/0002-quilt.mod.json.md
@@ -273,11 +273,41 @@ One or more paths to a square .PNG file. If an object is provided, the keys must
 ```
 
 ## The `mixin` field
-| Type         | Required |
-|--------------|----------|
-| Array/String | False    |
+| Type                | Required |
+|---------------------|----------|
+| Array/String/Object | False    |
 
-A single or array of paths to mixin configuration files relative to the root of the mod JAR.
+A single or array of mixin configuration entries. 
+
+An entry can either be an object or a string. 
+The mixin configuration entry object has one required field `config` and an optional field `environment`. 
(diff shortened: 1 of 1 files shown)


ix0rai (approved):

Ampflower:
Doesn't Mixin already natively support this via `client` and `server`?
Curious, why duplicate the work that already exists?

FirstMegaGame4:
the issue here is that quilt loader doesn't parse sided mixins in qmj

OroArmor:
This adds support for something already in Fabric Loader. The main use for this feature is for split-source mods, as this allows the client mixins to have their own separate mixin config. I could see a small use case for a mixin that has a different effect on the server vs the client.

AlexIIL (approved):

AlexIIL:
Final comment period ends in 3 days on January 16th.

MattiDragon:
There seems to be a stray "Providing" at the end of the mixin section. Is that intentional?

specification/0002-quilt.mod.json.md
```diff
@@ -273,11 +273,41 @@ One or more paths to a square .PNG file. If an object is provided, the keys must
...
+```json
+"mixin": [
+   {
+      "config": "modid_client.mixins.json",
+      "environment": "client"
+   },
+   "modid.mixins.json"
+]
+```
+Providing
```
OroArmor, line 310:
```suggestion
```

OroArmor:
> There seems to be a stray "Providing" at the end of the mixin section. Is that intentional?

Thanks

Pyrofab (approved):

Laxystem (approved):

TibiNonEst (approved):

